PR #2, opened 2017-06-13
[WIP] Adding contact methods to user resource

mattyjones:


mattyjones:
If anyone wants to jump in an offer a suggestion. I am trying add a list of contact methods to a user. I am making some progress but I am crashing terraform due to trying and failing to do a proper type assertion [here](https://github.com/mattyjones/terraform-provider-pagerduty/blob/master/pagerduty/resource_pagerduty_user.go#L125).

[crash.txt](https://github.com/terraform-providers/terraform-provider-pagerduty/files/1073011/crash.txt)

[users.txt](https://github.com/terraform-providers/terraform-provider-pagerduty/files/1073014/users.txt)

I am building and running this against the terraform master branch.

Thanks!

heimweh:
Hi @mattyjones, thanks for submitting this PR! 
managing contact methods for users is a very welcomed addition to the PagerDuty provider :)

It seems like there's a separate endpoint for managing contact methods for a user (although it seems like you can get some read-only information via the `User`)

What I would suggest then is to rather add a new resource for managing contact methods (so they can have their own lifecycle), e.g:

```hcl
resource "pagerduty_user" "foo" {
  email = "foo@bar.com"
  ...
}

resource "pagerduty_user_contact_method" "foo_email" {
  user_id = "${pagerduty_user.foo.id}"
  ...
}
```

I've added support for this in the client-library like so:
```go
// List
resp, _, err := client.ListContactMethods(userID)
if err != nil {
	return err
}

for _, cm := range resp.ContactMethods {
	fmt.Printf("Contact method ID: %s", cm.ID)
}

// Create
cm, _, err := client.Users.CreateContactMethod(userID, &pagerduty.ContactMethod{Type: "email_contact_method", Address: "foo@bar.com"})
if err != nil {
	return err
}

// Read
cm, _, err := client.Users.GetContactMethod(userID, contactMethodID)
if err != nil {
	return err
}

// Update
cm, _, err := client.Users.UpdateContactMethod(userID, contactMethodID, &pagerduty.ContactMethod{Type: "email_contact_method", Address: "foo@bar.com"})
if err != nil {
	return err
}

// Delete
if _, err := client.Users.DeleteContactMethod(userID, contactMethodID); err != nil {
	return err
}
```

By splitting this out to a separate resource, you also won't have to handle nested structures (I think) thus eliminating the issue you're seeing with the type assertion :) let me know what you think and if you need any help implementing this I'd be happy to help out, otherwise it's possible to handle nested structure like this by having a `flatten` & `expand` `func()` which we use quite heavily in the other resources (like `pagerduty_schedule`).

Nowaker:
Great work here! :+1:  Can't wait to see this finalized.

PR #14, opened 2017-06-30
docs: add undocumented attributes

nicolai86:


stack72:
Thanks @nicolai86 :) You Rock!

PR #30, opened 2017-08-23
provider: handle not found errors better

heimweh:
This should handle most, if not all resource not found errors returned by the PagerDuty API (https://v2.developer.pagerduty.com/docs/errors#pagerduty-error-codes).

### DIFF
--- a/pagerduty/provider.go
+++ b/pagerduty/provider.go
@@ -3,7 +3,6 @@ package pagerduty
 import (
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


heimweh:
@radeksimko checking for a `404` status code in the response seems to work quite well across the various resources and is a way cleaner solution, so let's go with that 🙂 many thanks!

radeksimko:
LGTM, :shipit:

Issue #34, opened 2017-08-23
Permadiff on schedule start time


2rs2ts:
### Terraform Version
0.10.2

(This was not an issue on 0.9.10. It became an issue after upgrading.)

### Affected Resource(s)
- pagerduty_schedule

### Terraform Configuration Files
```hcl
resource "pagerduty_schedule" "our_primary_on_call" {
  name      = "Our Primary On-call"
  time_zone = "America/Chicago"

  layer {
    start                        = "2017-08-08T17:16:03-05:00" # Rotation flips at 11AM Mon, Chi time
    rotation_virtual_start       = "2017-07-31T11:00:00-05:00"
    rotation_turn_length_seconds = 604800                      # 1 week

    users = ["${data.pagerduty_user.on_call_users.*.id}"]
  }
}
```

### Debug Output
I can provide an isolated example and produce debug output if necessary but would rather not have to do so and then also redact our private credentials.

### Expected Behavior
No diff should be produced after applying this once

### Actual Behavior
Subsequent plans, applies show a diff:

```
  ~ module.pagerduty.pagerduty_schedule.our_primary_on_call
      layer.0.start: "2017-08-23T15:09:35-05:00" => "2017-08-08T17:16:03-05:00"
```

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

heimweh:
Hi @2rs2ts, thanks for reporting this issue.

I managed to reproduce this issue, and I've published a fix here: #35. 

I'm very sorry about this and I hope this fix should resolve the issue you're currently experiencing.

Best,
Alexander

pepov:
Hi, this is great! When can we expect the next release?

2rs2ts:
Closing because the fix was merged.

2rs2ts:
But yes please release the fix. It's been almost 2 months since the fix was merged.

joseluis-fw:
@2rs2ts Any plan to include this in a new release?

2rs2ts:
@joseluis-fw I'm not a maintainer, so the idea has not crossed my mind :)

joseluis-fw:
@heimweh Any plan to include cut a new release?

PR #35, opened 2017-08-23
r/pagerduty_schedule: Don't read back `start` value for schedule layers.

heimweh:
This PR fixes a bug found in `pagerduty_schedule` (described in #34) related to `start`.

The PagerDuty API always returns a new `start` value, 
which in turn causes Terraform to get a diff on every plan. 

This PR should resolve that, by ignoring whatever value we get back from the API.

### DIFF
--- a/pagerduty/resource_pagerduty_schedule.go
+++ b/pagerduty/resource_pagerduty_schedule.go
@@ -176,6 +176,19 @@ func resourcePagerDutyScheduleRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("time_zone", schedule.TimeZone)
 	d.Set("description", schedule.Description)
 
+	// Here we override whatever `start` value we get back from the API
+	// and use what's in the configuration. This is to prevent a diff issue
+	// because we always get back a new `start` value from the PagerDuty API.
+	for _, sl := range schedule.ScheduleLayers {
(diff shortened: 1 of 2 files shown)


2rs2ts:
👏 thanks for the super quick response on this issue!

2rs2ts:
Is there something holding this back from being merged?

2rs2ts:
The PagerDuty API doesn't really document the resources it exposes or any of their quirks, just the generic API methods available. https://v2.developer.pagerduty.com/docs/endpoints

I poked around the API and I believe it's not that the API always returns a new value, but rather, the start time is getting updated every time terraform updates the schedule. It may actually be that there's a way to prevent that from happening, but I'm not sure right now.

heimweh:
@2rs2ts - Your observation is correct. The `start` time when updating a schedule layer will always be the time when the schedule layer was last updated, also confirmed this with the support team at PagerDuty.

@catsby - Unfortunately this behavior isn't documented anywhere, but I made an effort describing the behavior in the resource documentation. I don't think the updated value we get back from the API is of any real value, but it would in that case make more sense to store that updated value under a different name such as: `last_updated`. Would you mind taking a look again? many thanks! 🙇

Nowaker:
Nice @heimweh!

PR #36, opened 2017-08-23
d/pagerduty_service_integration: update list of arguments

heimweh:
This PR adds `integration_email` and `integration_key` as optional arguments.


sebradloff:
Can anyone tell me if `integration_key` option actually does anything? Based on what I see it doesn't look like there's a way it gets set in the pagerduty api anywhere. I try setting it like so:
```
resource "pagerduty_service_integration" "new_relic" {
  name                  = "NR to PD"
  service               = "${pagerduty_service.admin.id}"
  vendor               = "${data.pagerduty_vendor.new_relic.id}"
  integration_key = "unique-integration-key"
}
```

The main reason for injecting the integration_key is because I need it for the configuration of New Relic alert policy notification channel. If there was a data source for service integrations with integration_key as an attribute, this could solve the problem I'm running into.

The pattern we have is 1 New Relic alert policy per application but we can have many alert policies to 1 Pager Duty service.

PR #55, opened 2018-01-04
d/pagerduty_vendor: Match whole string and fallback to partial matching

### LINKED ISSUE #54: Vendor Data Source can't find Sentry
The following configuration doesn't find the Vendor "Sentry," but instead "IPSentry," obliging me to instead put the Sentry vendor's id in a local.

```
data "pagerduty_vendor" "sentry" {
  name = "Sentry"
}
```

I haven't looked at the code, but I have looked at the PagerDuty API.  Running `curl -X GET --header 'Accept: application/vnd.pagerduty+json;version=2' --header 'Authorization: Token token=<your token>' 'https://api.pagerduty.com/vendors?query=Sentry'` gives back both vendors, with IPSentry first.
```
{
  "vendors": [
    {
      "id": "PY3OLY6",
      "type": "vendor",
      "summary": "IPSentry",
      "self": "https://api.pagerduty.com/vendors/PY3OLY6",
      "html_url": null,
      "name": "IPSentry",
      "website_url": "http://ipsentry.com/",
      "long_name": "IPSentry",
      "logo_url": "",
      "thumbnail_url": "",
      "description": "IPSentry is a Windows based network monitoring software package used by thousands of information system specialists, system administrators, and IT solution providers around the world.",
      "integration_guide_url": "https://www.pagerduty.com/docs/guides/ipsentry-integration-guide/",
      "connectable": false,
      "generic_service_type": "api",
      "alert_creation_default": "create_alerts_and_incidents",
      "alert_creation_editable": true,
      "is_pd_cef": false
    },
    {
      "id": "PKG4M95",
      "type": "vendor",
      "summary": "Sentry",
      "self": "https://api.pagerduty.com/vendors/PKG4M95",
      "html_url": null,
      "name": "Sentry",
      "website_url": "https://getsentry.com/welcome/",
      "long_name": "Sentry",
      "logo_url": "https://s3.amazonaws.com/pdpartner/sentry_large.png",
      "thumbnail_url": "https://s3.amazonaws.com/pdpartner/sentry_thumb.png",
      "description": "Sentry provides real-time crash reporting for your web apps, mobile apps, and games.",
      "integration_guide_url": "https://www.pagerduty.com/docs/guides/sentry-integration-guide/",
      "connectable": false,
      "generic_service_type": "api",
      "alert_creation_default": "create_alerts_and_incidents",
      "alert_creation_editable": true,
      "is_pd_cef": false
    }
  ],
  "query": "Sentry",
  "limit": 25,
  "offset": 0,
  "total": null,
  "more": false
}
```

I think it'd be a good idea to find the object with the matching `name` field in the array, rather than just taking the first one.  This would also require users to specify the full, exact name of the vendor in the data source's `name`, which is what I assumed was the case from the start; as it stands, I get the same results using `"Sen"`.

heimweh:
Fixes #54 

This PR solves an issue where a vendor search for: "Sentry" would return "IPSentry" instead of "Sentry". To fix this we match the whole string instead.

### DIFF
--- a/pagerduty/data_source_pagerduty_vendor.go
+++ b/pagerduty/data_source_pagerduty_vendor.go
@@ -49,12 +49,23 @@ func dataSourcePagerDutyVendorRead(d *schema.ResourceData, meta interface{}) err
 
 	var found *pagerduty.Vendor
 
-	r := regexp.MustCompile("(?i)" + searchName)
+	er := regexp.MustCompile(fmt.Sprintf("^(?i)%s$", searchName))
 
 	for _, vendor := range resp.Vendors {
-		if r.MatchString(vendor.Name) {
+		if er.MatchString(vendor.Name) {
 			found = vendor
 			break
+		}
+	}
+
+	// We didn't find an exact match, so let's fallback to partial matching.
+	if found == nil {
+		pr := regexp.MustCompile("(?i)" + searchName)
+		for _, vendor := range resp.Vendors {
+			if pr.MatchString(vendor.Name) {
+				found = vendor
+				break
+			}
 		}
 	}
 
--- a/pagerduty/data_source_pagerduty_vendor_test.go
+++ b/pagerduty/data_source_pagerduty_vendor_test.go
@@ -1,14 +1,13 @@
 package pagerduty
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
-	"github.com/hashicorp/terraform/terraform"
 )
 
 func TestAccDataSourcePagerDutyVendor_Basic(t *testing.T) {
+	dataSourceName := "data.pagerduty_vendor.foo"
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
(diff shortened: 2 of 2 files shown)


yeungda-rea:
Hello @heimweh! We also experienced issue #54. Our first solution was to do an exact match too. But this _might_ break backwards compatibility with users who expect to be able to provide partial matches. A more compatible approach might be to prefer the exact match but still support the partial match:

```
...
var found *pagerduty.Vendor = findVendor(resp.Vendors, searchName)
...

func findVendor(vendors []*pagerduty.Vendor, searchName string) (*pagerduty.Vendor) {
	r := regexp.MustCompile("(?i)" + searchName)
	var closeMatch *pagerduty.Vendor
  for _, vendor := range vendors {
		if searchName == vendor.Name {
			return vendor
		}
		if closeMatch == nil && r.MatchString(vendor.Name) {
			closeMatch = vendor
		}
	}

	return closeMatch
}
```

heimweh:
Hi @yeungda-rea, thanks for your input. 
I was actually about to tag this as a breaking change, but I totally agree with you.
It's better to fallback to partial matching if the exact match fails, to avoid a breaking change.

Updated the tests, so we should now be able to cover exact (default) and partial matching (fallback).

yeungda-rea:
This implementation looks like it would suit our needs. Looking forward to it being merged so we can stop carrying a patch for bug #54.

PR #60, opened 2018-01-29
Don't re-create services if support hours or scheduled actions change

### LINKED ISSUE #53: Changing a PagerDuty service's support hours or scheduled actions recreates the service
Changing a PagerDuty service's support hours or scheduled actions recreates the service.

This appears to be intentional as the [docs](https://www.terraform.io/docs/providers/pagerduty/r/service.html) state
> Changes to support_hours necessitate re-creating the service resource.

and

> Changes [to scheduled actions] necessitate re-createing the service resource.

Re-creating a service under these circumstances causes us problems:
- we need to update our systems to post alerts to a new endpoint
- we lose alert history

It appears that this behaviour is controlled through hard-coded `true` values for [support hours](https://github.com/terraform-providers/terraform-provider-pagerduty/blob/master/pagerduty/resource_pagerduty_service.go#L120) and [scheduled actions](https://github.com/terraform-providers/terraform-provider-pagerduty/blob/master/pagerduty/resource_pagerduty_service.go#L151). If I was to change these to `false`, what, if anything would break? My initial testing indicates it seems okay when adding support hours or scheduled actions entities to a service, updating them, or removing them from a service.

This is currently a blocker for us so I'm more than happy to put together a pull request if you can point me in the right direction.

mbelekar:
We have added a fix so that services don't get recreated when we change scheduled action or support hours. Fixes #53

### DIFF
--- a/pagerduty/resource_pagerduty_service.go
+++ b/pagerduty/resource_pagerduty_service.go
@@ -120,7 +120,7 @@ func resourcePagerDutyService() *schema.Resource {
 				Optional: true,
 				MaxItems: 1,
 				MinItems: 1,
-				ForceNew: true,
+				ForceNew: false,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"type": {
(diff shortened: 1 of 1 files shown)


heimweh:
Hi @mbelekar, thank you so much for this contribution. 
This change looks good to me at a first glance. Would you mind updating the documentation so

heimweh:
Hi @mbelekar, thank you for this contribution! 

This change looks good to me at a first glance and just confirmed that this is possible without having to re-create the resource. Would you mind updating the documentation for this resource so it reflects the changes, also it would be nice if we could update the acc tests as well so we have that covered.

Many thanks!

Alexander

mbelekar:
Created PR #67 which has acceptance tests. Closing this PR

PR #69, opened 2018-02-25
Enable extensions in the provider

mbelekar:
This PR contains following changes

1. Updated client version so that it supports extensions
2. Added data source and acceptance test for extension schema
3. Added resource 'extension' and acceptance tests
4. Documentation on extensions

Test results:
```
➜  terraform-provider-pagerduty git:(extensions) ✗ TF_ACC=1 go test github.com/terraform-providers/terraform-provider-pagerduty/pagerduty -v -run=TestAccDataSourcePagerDutyExtensionSchema_Basic -timeout 120m
=== RUN   TestAccDataSourcePagerDutyExtensionSchema_Basic
--- PASS: TestAccDataSourcePagerDutyExtensionSchema_Basic (5.67s)
PASS
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	5.740s

➜  terraform-provider-pagerduty git:(extensions) ✗ TF_ACC=1 go test github.com/terraform-providers/terraform-provider-pagerduty/pagerduty -v -run=TestAccPagerDutyExtension_Basic -timeout 120m
--- PASS: TestAccPagerDutyExtension_Basic (22.68s)
PASS
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	22.759s
```

Please let me know if any changes are needed.

### DIFF
--- a/pagerduty/data_source_pagerduty_extension_schema.go
+++ b/pagerduty/data_source_pagerduty_extension_schema.go
@@ -50,7 +50,7 @@ func dataSourcePagerDutyExtensionSchemaRead(d *schema.ResourceData, meta interfa
 	}
 
 	d.SetId(found.ID)
-	d.Set("label", found.Label)
+	d.Set("name", found.Label)
 	d.Set("type", found.Type)
 
 	return nil
--- a/pagerduty/data_source_pagerduty_extension_schema.go
+++ b/pagerduty/data_source_pagerduty_extension_schema.go
@@ -1,0 +1,75 @@
+package pagerduty
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func dataSourcePagerDutyExtensionSchema() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePagerDutyExtensionSchemaRead,
+
+		Schema: map[string]*schema.Schema{
+			"name_regex": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Removed:  "Use `name` instead. This attribute will be removed in a future version",
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourcePagerDutyExtensionSchemaRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*pagerduty.Client)
+
+	log.Printf("[INFO] Reading PagerDuty Extension Schema")
+
+	searchName := d.Get("name").(string)
+
+	resp, _, err := client.ExtensionSchemas.List()
+	if err != nil {
+		return err
+	}
+
+	var found *pagerduty.ExtensionSchema = findExtensionSchema(resp.ExtensionSchemas, searchName)
+
+	if found == nil {
+		return fmt.Errorf("Unable to locate any extension schema with the name: %s", searchName)
+	}
+
+	d.SetId(found.ID)
+	d.Set("label", found.Label)
+	d.Set("type", found.Type)
+
+	return nil
+}
+
+func findExtensionSchema(extensionSchemas []*pagerduty.ExtensionSchema, searchName string) *pagerduty.ExtensionSchema {
+	r := regexp.MustCompile("(?i)" + searchName)
+	var closeMatch *pagerduty.ExtensionSchema
+	for _, extensionSchema := range extensionSchemas {
+		if searchName == extensionSchema.Label {
+			return extensionSchema
+		}
+		if r.MatchString(extensionSchema.Label) {
+			closeMatch = extensionSchema
+		}
+	}
+
+	if closeMatch != nil {
+		return closeMatch
+	}
+	return nil
+}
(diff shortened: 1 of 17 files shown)


pagerduty/data_source_pagerduty_extension_schema.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func dataSourcePagerDutyExtensionSchema() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePagerDutyExtensionSchemaRead,
+
+		Schema: map[string]*schema.Schema{
+			"name_regex": {
```
catsby: If I had to guess, I would think`name_regex` here was copied from somewhere? Perhaps PagerDutyVendor data source, where I also see it. It should be removed here as it's not used.

pagerduty/resource_pagerduty_extension.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func resourcePagerDutyExtensionCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*pagerduty.Client)
+
+	Extension := buildExtensionStruct(d)
+
+	log.Printf("[INFO] Creating PagerDuty extension %s", Extension.Name)
+
+	Extension, _, err := client.Extensions.Create(Extension)
```
catsby: Nitpick: I believe lowercase variable names are more idiomatic Go here: `Extension` -> `extension`.

pagerduty/resource_pagerduty_extension.go
```diff
@@ -0,0 +1,139 @@
...
+	client := meta.(*pagerduty.Client)
+
+	log.Printf("[INFO] Reading PagerDuty extension %s", d.Id())
+
+	extension, _, err := client.Extensions.Get(d.Id())
+	if err != nil {
+		return handleNotFoundError(err, d)
+	}
+
+	d.Set("name", extension.Name)
```
catsby: `name` appears to be the only attribute that is read from the API and updated in state here, unless I'm missing something? All possible attributes in the schema should be read from the API and updated in state.

pagerduty/resource_pagerduty_extension.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+func resourcePagerDutyExtensionDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*pagerduty.Client)
+
+	log.Printf("[INFO] Deleting PagerDuty extension %s", d.Id())
+
+	if _, err := client.Extensions.Delete(d.Id()); err != nil {
+		return err
```
catsby: Please check the error code here. If it's equivalent to a `404 Not Found` error, it's better to log the error and indicate that because it wasn't found, Terraform is removing the resource from state. 

Ex:

```go
if _, err := client.Extensions.Delete(d.Id()); err != nil {
  if err == client.404Error { // pseudo code 
    log.Printf("[WARN] Extension (%s) not found, removing from state" , d.Id())
    return nil // returning nil from DELETE will remove item from state
  }
  return err
}
```

This allows Terraform to gracefully handle the removal of resources that have been destroyed outside of Terraform.

pagerduty/resource_pagerduty_extension.go
```diff
@@ -0,0 +1,139 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"endpoint_url": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"extension_objects": {
+				Type:     schema.TypeList,
```
catsby: Are `extension_objects` ordered? If the ordering for these objects is relevant, then `TypeList` is correct and we need to ensure we're reading `extension_objects` and storing them in state as the API has them. 

If ordering is not relevant, consider using `TypeSet` instead.

pagerduty/resource_pagerduty_extension_test.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+	}
+
+	return nil
+}
+
+func TestAccPagerDutyExtension_Basic(t *testing.T) {
+	extension_name := fmt.Sprintf("tf-%s", acctest.RandString(5))
+	extension_name_updated := fmt.Sprintf("tf-%s", acctest.RandString(5))
+	name := fmt.Sprintf("tf-%s", acctest.RandString(5))
```
catsby: Nit: We have some helpers if you're interested:

- https://github.com/hashicorp/terraform/blob/7c6072c2a00b54dd3a7d0a19436052b2e4b553ab/helper/resource/id.go#L1

```go
extension_name := resource.PrefixedUniqueId("tf")
```

website/docs/r/extension.html.markdown
```diff
@@ -0,0 +1,81 @@
...
+layout: "pagerduty"
+page_title: "PagerDuty: pagerduty_extension"
+sidebar_current: "docs-pagerduty-resource-extension"
+description: |-
+  Creates and manages a service extension in PagerDuty.
+---
+
+# pagerduty\_extension
+
+A [extension](https://v2.developer.pagerduty.com/v2/page/api-reference#!/Extensions/post_extensions) can be associated with a service.
```
catsby: Nit: `A extension ` -> `An extension `

pagerduty/resource_pagerduty_extension.go
mbelekar: I am setting additional atrributes now.

mbelekar:
Thanks @catsby for the feedback. I have made changes accordingly.
Cheers!

PR #70, opened 2018-03-08
Update `pagerduty_extension` to return READ() for both CREATE and UPDATE

catsby:
Update the `pagerduty_extension` resource to return the results of it's `READ` method at the end of both `CREATE` and `UPDATE`. This pattern is used to detect drift in the event that data was sent to the API without error, but resulted in different values being set (typically from API changing data, etc).

### DIFF
--- a/pagerduty/resource_pagerduty_extension.go
+++ b/pagerduty/resource_pagerduty_extension.go
@@ -4,6 +4,7 @@ import (
 	"log"
 
 	"fmt"
+
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/heimweh/go-pagerduty/pagerduty"
 )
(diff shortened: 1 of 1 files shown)


catsby:
Hey @mbelekar here are the changes I missed in my first review of #69 , basically a best-practice pattern to update state with the latest data from the API.

catsby:
Tests pass so I'm going to merge (ﾉ^_^)ﾉ

```
$ make testacc TESTARGS='-run=TestAccPagerDutyExtension' 
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccPagerDutyExtension -timeout 120m
?       github.com/terraform-providers/terraform-provider-pagerduty     [no test files]
=== RUN   TestAccPagerDutyExtension_import
--- PASS: TestAccPagerDutyExtension_import (18.39s)
--- PASS: TestAccPagerDutyExtension_Basic (19.36s)
PASS
ok      github.com/terraform-providers/terraform-provider-pagerduty/pagerduty   37.776s
```

PR #71, opened 2018-03-08
Error check extension_object read in PagerDuty Extensions; use flatten method

catsby:
Hello! #69 added `pagerduty_extension` resource, and only too late did I notice that we aren't checking the error when setting the non-scalar value `extension_object`. Sure enough, it was returning an error due to incompatible format:

```
2018-03-08T12:09:05.781-0600 [DEBUG] plugin.terraform-provider-pagerduty: 2018/03/08 12:09:05 [WARN] error setting extension_objects: extension_objects.0: '' expected type 'string', got unconvertible type '*pagerduty.ServiceReference'
```

This means reading from the API would not detect drift, should something have changed outside of Terraform. `extension_object` is a `ForceNew` attribute, so perhaps that's not possible anyway (the extension we know may be recreated with a new ID if changed externally). 

Either way, I add check the error here and add a `flattenExtensionObjects` method.


catsby:
cc @mbelekar if you're interested

PR #72, opened 2018-03-10
Create Dockerfile to enable easier local development

sebradloff:
Created a Dockerfile loaded with Go 1.8 and Terraform, figured it would be easier for local development. Might be worth thinking about a test API token for the acceptance tests, mainly due to the fact that some people might not have account features necessary for running these tests.


sebradloff:
Can I please get 👀 or any feedback? Happy to change the PR to be more in line with the development of this plugin.

PR #92, opened 2018-08-02
Docs: add valid user color options

tdmalone:
I couldn't find these documented anywhere in the PagerDuty API or elsewhere, so I generated this list directly from the system names used in the color dropdown in the PagerDuty UI.


tdmalone:
@heimweh Of course, apologies for the delay and sorry I didn't notice all the dupes when I first copied it out! I've sorted and deduped now.

PR #122, opened 2019-03-28
Add support for schedule overrides

acrewdson:
(Relates to https://github.com/terraform-providers/terraform-provider-pagerduty/issues/112)

This works but is still WIP and I'd need to improve and expand the tests. Before I do that I want to make sure the maintainers think this is an OK approach. Specifically I'd like to know if it seems right to add this as a new resource, rather than as an argument to `pagerduty_schedule`. What I've got in this PR looks like:

```
resource "pagerduty_schedule_override" "my_override" {
  user     = "${pagerduty_user.my_user.id}"
  start    = "2019-04-01T20:00:00-04:00"
  end      = "2019-04-02T20:00:00-04:00"
  schedule = "${pagerduty_schedule.my_schedule.id}"
}
```

As opposed to:

```
resource "pagerduty_schedule" "foo" {
  name      = "Daily Engineering Rotation"
  time_zone = "America/New_York"

  layer {
    name                         = "Night Shift"
    start                        = "2015-11-06T20:00:00-05:00"
    rotation_virtual_start       = "2015-11-06T20:00:00-05:00"
    rotation_turn_length_seconds = 86400
    users                        = ["${pagerduty_user.foo.id}"]
  }

  override {
    user = ...
    etc
  }
}
```

Making overrides their own resource seems to match the way service integration resources are distinct from service resources, and also aligns with the way the PagerDuty REST API does stuff. But I'd certainly defer to the maintainers here as to what makes the most sense. Thank you for considering this change!

### DIFF
--- a/pagerduty/provider.go
+++ b/pagerduty/provider.go
@@ -47,6 +47,7 @@ func Provider() terraform.ResourceProvider {
 			"pagerduty_user":                resourcePagerDutyUser(),
 			"pagerduty_user_contact_method": resourcePagerDutyUserContactMethod(),
 			"pagerduty_extension":           resourcePagerDutyExtension(),
+			"pagerduty_schedule_override":   resourcePagerDutyScheduleOverride(),
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/pagerduty/resource_pagerduty_schedule_override.go
+++ b/pagerduty/resource_pagerduty_schedule_override.go
@@ -1,0 +1,123 @@
+package pagerduty
+
+import (
+	"errors"
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func resourcePagerDutyScheduleOverride() *schema.Resource {
+	return &schema.Resource{
+		Create: resourcePagerDutyScheduleOverrideCreate,
+		Read:   resourcePagerDutyScheduleOverrideRead,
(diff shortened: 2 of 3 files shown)


heimweh:
Hi @acrewdson ,

Thank you so much for this draft. 

Extremely sorry for the late reply but I totally agree that this should be a separate resource.

Thank you so much once again!
This will be a nice addition to the PagerDuty provider.

Best,

Alexander

acrewdson:
@heimweh No problem, thanks for the feedback. In that case, I'll work on expanding the tests and converting this to a real PR. Thank you!

stmcallister:
@acrewdson Thanks for doing this! Just wanted to check in and see the status on this feature. Is this something that is almost done, or would you like a hand on it?

acrewdson:
Hey @stmcallister 👋 

I think we'd need to add more extensive tests to gain confidence that what's here is on the right track. Way back when I first worked on this I added some `TODO`s that would be a starting point for expanding tests: https://github.com/terraform-providers/terraform-provider-pagerduty/pull/122/files#diff-952d3eb6d7501b3cf03bf2cd6fe16374R291

I may have a chance to return to those in the next week or two, but you're also welcome to dig in if you'd like. Any help is certainly appreciated. Thanks!

acrewdson:
Probably before we spend more time on this PR, it could be good to get some signal from maintainers (@heimweh ?) that there would be eagerness to merge this if/when tests were expanded. I might be hesitant to devote more effort until then, but I'm speaking only for myself here for sure.

mdevens:
This is something I could use - hoping this feature does get added at some point.

PR #135, opened 2019-05-29
Remove invalid role types

luis-silva:
According to [the documentation](https://support.pagerduty.com/docs/advanced-permissions#section-base-roles), pagerduty no longer supports the `team_responder` role. It was replaced by the `observer` role. This creates a constant diff when running terraform since the request is issued with the legacy type but pagerduty will internally convert it to the new type.

I have also removed the owner role from the validation function as it was being ignored by the provider anyway and not sent with the request to the Pagerduty API, causing constant diffs on that role as well.

### DIFF
--- a/pagerduty/resource_pagerduty_user.go
+++ b/pagerduty/resource_pagerduty_user.go
@@ -37,13 +37,6 @@ func resourcePagerDutyUser() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "user",
-				ValidateFunc: validateValueFunc([]string{
-					"admin",
-					"limited_user",
-					"read_only_user",
-					"observer",
-					"user",
-				}),
 			},
 
 			"job_title": {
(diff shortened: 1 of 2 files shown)


luis-silva:
@heimweh @gordonbondon Can any of you take a quick look at this?

gordonbondon:
@luis-silva hi! I'm not the maintainer of this provider, so, unfortunately, I won't be able to help you get this merged.

luis-silva:
@heimweh Any chance you can review this?

heimweh:
Hi @luis-silva, 

thank you so much for this contribution and sorry for the delayed response :pray:

This looks good! Though I'd rather want to remove our static validation and let PagerDuty return an error and/or guide the user to the right value since the list of roles might change again. Would you mind removing the `ValidateFunc` altogether? 🙂 

Thank you so much once again and so sorry for the delay on this one.

Best,

Alexander

luis-silva:
@heimweh It's even better that way. I did think about that but didn't want this PR to make more changes than required and risk it not being accepted.

I have now remove the validator function altogether.

luis-silva:
@heimweh Thanks for merging it. Can you publish a new release so these changes are available in terraform?

PR #146, opened 2019-08-02
resource/pagerduty_user: deprecate teams attribute

heimweh:
Here we're deprecating the `teams` attribute.
Teams can be added/removed using the dedicated resource `pagerduty_team_membership`.

This attribute has been quite tricky to maintain and is causing a lot of weird dependency conflicts and failing tests. In fact, team membership can be inherited from an escalation policy causing issues with this attribute. 

Relates #147

This change is also more in line with the PagerDuty API.

Before:

```hcl
resource "pagerduty_team" "foo" {
  name = "foo"
}

resource "pagerduty_user" "foo" {
  name  = "foo"
  email = "foo@foo.com"
  teams = ["${pagerduty_team.foo.id}"]
}
```

After:
```hcl
resource "pagerduty_team" "foo" {
  name = "foo"
}

resource "pagerduty_user" "foo" {
  name  = "foo"
  email = "foo@foo.com"
}

resource "pagerduty_team_membership" "foo" {
  user_id = "${pagerduty_user.foo.id}"
  team_id = "${pagerduty_team.foo.id}"
}
```

### DIFF
--- a/pagerduty/resource_pagerduty_team_membership_test.go
+++ b/pagerduty/resource_pagerduty_team_membership_test.go
@@ -81,7 +81,6 @@ func testAccCheckPagerDutyTeamMembershipConfig(user, team string) string {
 resource "pagerduty_user" "foo" {
   name = "%[1]v"
   email = "%[1]v@foo.com"
-  teams = ["${pagerduty_team.foo.id}"]
 }
 
 resource "pagerduty_team" "foo" {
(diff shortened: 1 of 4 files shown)


heimweh:
> This is a reasonable decision 👍
> 
> We had to deprecate some similar attributes in some other providers for very similar reason.
> 
> I reckon this would go out as minor release, and eventually we'd remove the field with `2.0.0` when the time comes, right?

That sounds like a good plan :ok_hand: Thank you so much @radeksimko :pray:

rlmartin:
When do you think this will be tagged and released?

rlmartin:
Or, more importantly, when will the next MAJOR version be released, per https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-removal ?

PR #150, opened 2019-08-10
Adding Event Rule Resource

stmcallister:
This PR adds support for the Event Rules endpoint in the PagerDuty API. This relates to [Issue 114](https://github.com/terraform-providers/terraform-provider-pagerduty/issues/114)

## Acceptance Test Results
```
TF_ACC=1 go test -run "TestAccPagerDutyEventRule_Basic" ./pagerduty -v -timeout 120m
=== RUN   TestAccPagerDutyEventRule_Basic
--- PASS: TestAccPagerDutyEventRule_Basic (9.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	10.906s


TF_ACC=1 go test -run "TestAccPagerDutyEventRule_import" ./pagerduty -v -timeout 120m
=== RUN   TestAccPagerDutyEventRule_import
--- PASS: TestAccPagerDutyEventRule_import (6.06s)
PASS
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	6.144s
```


pagerduty/resource_pagerduty_event_rule.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return eventRule
+}
+
+func expandString(v string) []interface{} {
+	var obj []interface{}
+	err := json.Unmarshal([]byte(v), &obj)
+
+	if err != nil {
+		log.Printf(string(err.Error()))
```
heimweh: What do you think about returning the error here if JSON unmarshal fails?

pagerduty/resource_pagerduty_event_rule.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	return obj
+}
+
+func flattenSlice(v []interface{}) string {
+	b, err := json.Marshal(v)
+	if err != nil {
+		log.Printf(string(err.Error()))
```
heimweh: Same as above, do we want to return the error here?

pagerduty/resource_pagerduty_event_rule.go
```diff
@@ -0,0 +1,145 @@
...
+func resourcePagerDutyEventRuleRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*pagerduty.Client)
+
+	log.Printf("[INFO] Reading PagerDuty event rule: %s", d.Id())
+
+	resp, _, err := client.EventRules.List()
+	if err != nil {
+		return err
+	}
+	for _, rule := range resp.EventRules {
```
heimweh: What do you think about doing something like:

```go
var found *pagerduty.EventRule

for _, rule := range resp.EventRules {
  if rule.ID == d.Id() {
    found = rule
    break
  }
}

if found == nil {
  d.SetId("")
}
...
d.Set("catch_all", rule.CatchAll)
...
```


That way you can set the attributes outside the loop 🙂

pagerduty/resource_pagerduty_event_rule.go
```diff
@@ -0,0 +1,145 @@
...
+		log.Printf("[DEBUG] Resp rule.ID: %s", rule.ID)
+		if rule.ID == d.Id() {
+			d.Set("action_json", flattenSlice(rule.Actions))
+			d.Set("condition_json", flattenSlice(rule.Condition))
+			d.Set("advanced_condition_json", flattenSlice(rule.AdvancedCondition))
+			d.Set("catch_all", rule.CatchAll)
+		}
+	}
+	// check if eventRule  not  found
+	if _, ok := d.GetOk("action_json"); !ok {
```
heimweh: Does this work? This calls `d.SetId("")` if the response code was a 404. 
I added an example above on how to clear it from the state if the event rule is missing from the list of event rules 🙂

pagerduty/resource_pagerduty_event_rule.go
stmcallister: For consistency, I've adopted the same strategy used in the [extension resource](https://github.com/terraform-providers/terraform-provider-pagerduty/blob/e9bf5a56146b00cb2fef5af92720b1156a5265d7/pagerduty/resource_pagerduty_extension.go#L196). Would that work?

pagerduty/resource_pagerduty_event_rule.go
stmcallister: This makes sense. I've adopted this idea in my next commit.

pagerduty/resource_pagerduty_event_rule.go
stmcallister: It appeared to have worked. 🤷‍♂ But, I've changed that logic to match what you were saying in the above comment.

PR #172, opened 2019-12-17
DOCS: fix formatting on pd_service parameter

gschaffer-cxn:
The possible value of intelligent was quoted instead of marked with backticks. Making sure it is the same as the value time and increase visibility in the text.


stmcallister:
LGTM! Thank you for the improvement!

PR #187, opened 2020-03-08
Add Teams To Service

bassrock:
This adds teams to a pagerduty service and should solve https://github.com/terraform-providers/terraform-provider-pagerduty/issues/109

However, there looks to be a bug in Pagerduty where you can't actually assign a team to a service, while their api supports the syntax here it does not do anything. 

I can also not find the page in Pagerduty's interface to associate a team with a service.

### DIFF
--- a/pagerduty/resource_pagerduty_service.go
+++ b/pagerduty/resource_pagerduty_service.go
@@ -198,6 +198,11 @@ func resourcePagerDutyService() *schema.Resource {
 					},
 				},
 			},
+			"teams": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
 		},
 	}
 }
@@ -264,6 +269,10 @@ func buildServiceStruct(d *schema.ResourceData) (*pagerduty.Service, error) {
 
 	if attr, ok := d.GetOk("support_hours"); ok {
 		service.SupportHours = expandSupportHours(attr)
+	}
+
+	if attr, ok := d.GetOk("teams"); ok {
+		service.Teams = expandTeams(attr)
 	}
 
 	return &service, nil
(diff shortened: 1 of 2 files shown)


wlonkly:
@bassrock I have a theory about the API behavior, but it might have to wait til Monday until I can confirm. In the web interface, though, you associate a team with an escalation policy, and an escalation policy with a service, so the team-to-service mapping is sort of transitive. 

(My theory is that you can _read_ a service's teams, but not _update_ them, thorugh the `/service` API endpoint. I could be completely wrong, though.)

I'll update here when I find out more (if no-one beats me to it!). I'll also pass on feedback to our product team about the problems you ran into.

bassrock:
@wlonkly Oh interesting. If thats the case then yea the Web UI is really mis-leading and the api docs as well. Thanks for getting back to me!!

wlonkly:
@bassrock what API docs were you using?

bassrock:
@wlonkly It was https://api-reference.pagerduty.com/#!/Services/get_services The request sample doesn't include teams but it has `teams` mentioned as a payload option.

wlonkly:
@bassrock Aha, ok, I see. `GET /service/:id` does return `teams` as a convenience (we do the dereference through the escalation policy for you) but if you look at the request sample for `PUT` or `POST` you'll see that there's no `teams` in there. For what it's worth, I had the same surprise until I got my head around the "service _has-a_ EP _has-a_ team" mental model.

(Feedback passed on to product and docs owners, too. 👍)

I suspect this means that `teams` would need to be exported attributes of the `pagerduty_service` resource and not resource arguments.

I should probably also note that I'm just a bystander here, no official involvement with this provider!

bassrock:
@wlonkly Yea that makes sense. I just verified the "service has-a EP has-a team" in the interface and that does work. So you are definitely right there. 

The appropriate action is to have pagerduty have a team export but not an input.


Thank you so much for helping understand that mental model. With that in mind it looks like the terraform resource already does what is needed!

stmcallister:
@bassrock if i'm reading this thread correctly, this PR is no longer needed. Is that correct?

bassrock:
@stmcallister Yep no longer needed.

PR #194, opened 2020-04-01
DOCS: Team membership role defaults to manager

tecnobrat:
This was added in https://github.com/terraform-providers/terraform-provider-pagerduty/pull/151, but the default team membership role is `manager`, update the docs to match.


stmcallister:
Good catch! Thank you!

PR #208, opened 2020-04-16
Implement retry logic in all read requests

mLupine:
Although the changes I made in #191 fixed the problem described in my previous PR,  we started encountering the same problems in read requests other than service and service integrations.

This PR implements the same retry logic to all data sources and resources. As previously, the changes are already field-tested in our organization 👍


stmcallister:
Thanks for submitting this! It's a much needed feature. Would you mind also adding a sleep on each retry? That way we give the API a break. Even just one second would be fine.

mLupine:
Hey @stmcallister! Added a delay before returning any retryable error, that should do the trick :)

stmcallister:
Well done! Thank you for doing that. i reduced each delay down to 2 seconds, and added the same retry logic to creating Event Rules so that their tests will finally pass. :) Appreciate the work!

PR #247, opened 2020-06-25
Add user contact method data source

mLupine:
The next step of _terraforming_ our PagerDuty configuration at @Nordcloud was to standardize and, well, create the Terraform configuration of on-call user notifications. However, during the process, we encountered a problem with one of the notification types — push notifications. Quite obviously, they cannot be configured via the API (so we can't create a resource for them and import it), but they can be _read_ from the API. This allows us to create a new data source for the contact methods and use it in notification rules.

And that's precisely what I've done in this pull request. It introduces a new data source - `pagerduty_user_contact_method`, along with an updated documentation and (passing!) tests 😉 .

As always, the changes are already live in our environment and we'll soon move it to production.


stmcallister (approved):
Thanks for adding this!! 🎉 Sorry for the delay in reviewing. 🤦

PR #252, opened 2020-07-16
Patch to allow services' status to be managed by the provider

qhua948:


jjm:
I was wondering if you have managed to confirm that services are created in an disabled state? Thing is I've got a long running support case where setting a state of `disabled` on a newly created service has no effect and it's just created in a `active` state (I'm told this is the correct behaviour).

From my understanding form our own tooling (not Terraform) is that the service would only end up being disabled after the a second apply, unless it confirms the service is in the correct state (which is the workaround I've reached).

qhua948:
> I was wondering if you have managed to confirm that services are created in an disabled state? Thing is I've got a long running support case where setting a state of `disabled` on a newly created service has no effect and it's just created in a `active` state (I'm told this is the correct behaviour).
> 
> From my understanding form our own tooling (not Terraform) is that the service would only end up being disabled after the a second apply, unless it confirms the service is in the correct state (which is the workaround I've reached).

I think I did test it with creating a `disabled` service. 

I have since lost access to pagerduty's API, would appreciate if you could test it out as well.

jjm:
I've not done much GoLang before, but my support case has been updated and the API docs been updated with the following

> If status is included in the request, it must have a value of active when creating a new service. If a different status is required, make a second request to update the service.

(taken from from https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1services/post)

They are looking at making a request with a status of disabled result in a 400 error. So this PR would beed to create the service as **active** and _then_ set the status to disabled.

stmcallister:
Thanks for adding this! I didn't even know about the status field until I saw this 😅 In general everything looks good. However, the test is failing with the following message.

Issue #236, opened 2020-05-27
pagerduty_service_dependency: does not support update


mhyzon:
It looks like the PD API doesn't support updating for the `pagerduty_service_dependency` resource.  Probably have to destroy & recreate instead.

### Terraform Version
```
$ terraform -v
Terraform v0.12.23
+ provider.pagerduty v1.7.1
```

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_service_dependency

### Debug Output

(I've obfuscated the IDs & names)

```
# module.business_services.pagerduty_service_dependency.technical_service_dependancy["XXXXX"] will be updated in-place
  ~ resource "pagerduty_service_dependency" "technical_service_dependancy" {
        id = "XXXXXXXXXXXXXX"

      ~ dependency {
            dependent_service {
                id   = "XXXXXXX"
                type = "business_service"
            }

          ~ supporting_service {
              ~ id   = "ABCDEF1" -> "ABCDEF2"
                type = "service"
            }
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.

...

Error: doesn't support update

  on ../../../modules/pagerduty/business_services/main.tf line 31, in resource "pagerduty_service_dependency" "technical_service_dependancy":
  31: resource "pagerduty_service_dependency" "technical_service_dependancy" {
```

### Expected Behavior
Assuming that Pagerduty API will not support an update on this object any time soon, terraform should destroy/create the resource.

### Actual Behavior
`Error: doesn't support update`

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1.  Change configuration so an existing pagerduty_service_dependency resource has a different supporting service
1.  `terraform apply`

mhyzon:
There might be an edge case here where the service dependency was removed on the PagerDuty side, but not removed from state (perhaps when the supporting service was removed? I might have done that manually.)

Playing around with the PD API directly, it was saying it couldn't find the service when I tried to dissociate manually. I removed it from state manually and new resource got created fine.

stmcallister:
Hi @mhyzon, yes, the underlying API for `service_dependencies` does not support updating. So, to make changes you'll need to destroy the dependency in your Terraform state and then create a new `service_dependency` resource. I've added a note to the `service_dependency` docs to help clarify this. Thanks for pointing out this wasn't clear!

0x91:
[...]

PR #271, opened 2020-10-01
resource/pagerduty_service_dependency: Update schema to disallow attribute updates

mlclmj:
👋 Hi there! I'm new to working with this provider, but wanted to take a stab at fixing #236 since our team ran into it today.

At the moment, though the resource schema for service dependencies marks the `dependency` attribute with the `forceNew` flag, updates to the dependent/supporting services still generate an "update diff" within Terraform which looks something like:

```hcl
  ~ resource "pagerduty_service_dependency" "glue" {
        id = "<removed>"

      ~ dependency {
          ~ dependent_service {
              ~ id   = "xxxxxx" -> "abcdefghi"
                type = "business_service"
            }

          ~ supporting_service {
              ~ id   = "xxxxxx" -> "bcdefghij"
                type = "service"
            }
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

This issue is detailed in #236, but disallowing this and forcing re-creation would be awesome since PagerDuty's API doesn't allow these associations to be updated.

Setting the `forceNew` flag on all of the schema attributes prevents Terraform from allowing any updates to the resource without re-creating it, a pattern that is followed [elsewhere in terraform](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lb.go#L103-L130)

### DIFF
--- a/pagerduty/resource_pagerduty_service_dependency.go
+++ b/pagerduty/resource_pagerduty_service_dependency.go
@@ -34,10 +34,12 @@ func resourcePagerDutyServiceDependency() *schema.Resource {
 									"id": {
 										Type:     schema.TypeString,
 										Required: true,
+										ForceNew: true,
 									},
 									"type": {
 										Type:     schema.TypeString,
 										Required: true,
+										ForceNew: true,
 									},
 								},
 							},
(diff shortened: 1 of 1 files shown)


stmcallister (approved):
Thank you for making this change! Makes sense. 🎉

PR #281, opened 2020-11-16
Make endpoint_url sensitive

mattcanty:
The endpoint URL query string contains API key


stmcallister (approved):
Good call. Thanks for submitting this! 🎉

PR #282, opened 2020-11-18
Fix issue with disabling a ruleset

### LINKED ISSUE #250: pagerduty_ruleset_rule - cannot set disabled=true once disabled=false was set
### Terraform Version
terraform: v0.12.8
terraform-provider-pagerduty: v1.7.3

### Affected Resource(s)
pagerduty_ruleset_rule 

```hcl
///////////////////////////////////////////////////////
// Example Rule
// This rule demonstrates a bug  when changing the value of 'disabled' from true to false
resource "pagerduty_ruleset_rule" "rule0" {
  position = 0
  ruleset  = "${pagerduty-ruleset.ruleset.id}"

  disabled = false /* Updating this from false to true does not work. Rule will remain disabled permanently.*/
  
  conditions {
    operator = "and"
    subconditions {
      operator = "matches"
      parameter {
        value = "(?i).*test.*"
        path  = "AlarmName"
      }
    }
  }

  actions {
    route {
      value = "${pagerduty-service.service.id}"
    }
  }
}
```

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
A rule that was deployed with `disabled=true` initially, should be able to be re-enabled by running Apply again after changing it to `disabled=false` 

### Actual Behavior
Rule remained Disabled.

### Steps to Reproduce
1. Set `disabled=true` on a rule
2. Run `terraform apply` 
3. Verify that the rule was created and is Disabled
4. Update Terraform script so that `disabled=false` on the same rule
5. Run `terraform apply`
6. Observe that the rule is still disabled.

### Important Factoids
N/A

### References
N/A

divbhasin:
Fixes #250. 

There are 2 main changes here:
- I added to the 2 existing ruleset_rule tests. Now, we first create a ruleset rule that has disabled=true, and then we update the same ruleset rule to disabled=false. This helps verify that we can go from true -> false, which was an issue before according to #250. We still get the verification of creating and updating ruleset rules that we had before.

- There was a change I made to the go-pagerduty client to enable this to work [here](https://github.com/heimweh/go-pagerduty/pull/45). Thus, I needed to vendor the go-pagerduty repo, which also grabbed some other updates to the client.


stmcallister (approved):
Looks good! thank you for submitting this fix!

PR #284, opened 2020-11-19
Make example show fully-qualified email address

origamislayer:
Bring example in-line with argument documentation lower on the page


stmcallister (approved):
Nice catch! Thank you!!! 🎉

Issue #307, opened 2021-02-28
routing_keys is still not returned by the pagerduty_ruleset data source


DoctorPolski:
### Terraform Version
0.14.7

### Affected Resource(s)
- `pagerduty_ruleset`

### Expected Behavior
`pagerduty_ruleset` data source should contain `routing_keys`

### Actual Behavior
Despite the recent update the ruleset data source still does not return `routing_keys`.

### Steps to Reproduce
1. terraform plan
2. "This object does not have an attribute named "routing_keys"."

### References
PR: https://github.com/PagerDuty/terraform-provider-pagerduty/pull/305

### Discussion
The function requires the `routing_keys` schema to be added in order for it to return them:

```
func dataSourcePagerDutyRuleset() *schema.Resource {
	return &schema.Resource{
		Read: dataSourcePagerDutyRulesetRead,

		Schema: map[string]*schema.Schema{
			"name": {
				Type:     schema.TypeString,
				Required: true,
			},
		},
	}
}
```

should be:

```
func dataSourcePagerDutyRuleset() *schema.Resource {
	return &schema.Resource{
		Read: dataSourcePagerDutyRulesetRead,

		Schema: map[string]*schema.Schema{
			"name": {
				Type:     schema.TypeString,
				Required: true,
			},
			"routing_keys": {
				Type:     schema.TypeList,
				Computed: true,
				Elem: &schema.Schema{
					Type: schema.TypeString,
				},
			},
		},
	}
}
```

When compiled and used locally the above works.

I originally requested this functionality last week directly via the PagerDuty helpdesk. I would be happy to submit this as a PR if given the required repo permissions.

Many thanks :)

stmcallister:
Thanks for submitting the PR!

stmcallister:
Going to close this issue. Please reopen if the issue still persists.

PR #312, opened 2021-03-04
Update to fix non-working ruleset data source which does not return routing_keys

DoctorPolski:
This commit is an extension of the following issue:
https://github.com/PagerDuty/terraform-provider-pagerduty/issues/307

It adds the required extra schema definition to fully enable the
ruleset data source to return routing_keys.


stmcallister (approved):
Doh! 🤦 That was my bad. Thank you for catching and fixing this!

PR #317, opened 2021-03-25
add details to event rules actions

tyurina:


stmcallister (approved):
Thanks for adding more details to the event rule docs!

PR #326, opened 2021-04-15
deps(go): use go1.16

chenrui333:


chenrui333:
cc @stmcallister
relates to #333

stmcallister (approved):
Thank you! 👍

PR #327, opened 2021-04-20
Fix error message for contact method import

asherf:
noted while importing that "user_id.contact_method_id" doesn't work but "user_id:contact_method_id" does.


asherf:
@stmcallister ping.

stmcallister (approved):
Good catch! Thanks for fixing this!

PR #340, opened 2021-05-26
Improve the `ruleset_rule` `time_frame` documentation

metavida:
Specifically, the previous description of how to choose a `start_time` value was misleading & did not explain how picking a correct value also involves the specified `timezone`

Also update the code example to demonstrate how to use [the `time_static` resource](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/static) for slightly more human-readable configuration. Best I could tell there is **not** a way to use other [built-in time functions like `formatdate`](https://www.terraform.io/docs/language/functions/formatdate.html) to create a unix epoch timestamp, which is why I picked the `time_static` resource.

A future improvement to the provider might include allowing users to specify `start_time` and other unix timestamp attributes using [RFC 3339](https://tools.ietf.org/html/rfc3339) "Date and Time format" syntax, but that improvement is out of the scope of this PR.

### DIFF
--- a/website/docs/r/ruleset_rule.html.markdown
+++ b/website/docs/r/ruleset_rule.html.markdown
@@ -41,9 +41,9 @@ resource "pagerduty_ruleset_rule" "foo" {
       # Every Tuesday, Thursday, & Saturday
       weekdays = [2, 4, 6]
       # Starting at 9:30am
-      start_time = "${time_static.eastern_time_at_0930.unix * 1000}"
+      start_time = time_static.eastern_time_at_0930.unix * 1000
       # Until 11:30am (2 hours later)
-      duration = "${2 * 60 * 60 * 1000}"
+      duration = 2 * 60 * 60 * 1000
       # in this timezone
       # (either EST or EDT depending on when your event arrives)
       timezone = "America/New_York"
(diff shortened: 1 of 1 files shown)


website/docs/r/ruleset_rule.html.markdown
```diff
@@ -23,34 +23,48 @@ resource "pagerduty_ruleset" "foo" {
...
     scheduled_weekly {
-	  weekdays = [3,7]
-	  timezone = "America/Los_Angeles"
-	  start_time = "1000000"
-	  duration = "3600000"
-	}
+      # Every Tuesday, Thursday, & Saturday
+      weekdays = [2, 4, 6]
+      # Starting at 9:30am
+      start_time = "${time_static.eastern_time_at_0930.unix * 1000}"
```
stmcallister, line 44: We're trying to standardize on the syntax introduced in Terraform v0.12 (i think). Please change this to match that notation as well. Essentially, just removing the double-quotes, dollar sign and curly-brackets. :) Something like this

`start_time = time_static.eastern_time_at_0930.unix * 1000`

website/docs/r/ruleset_rule.html.markdown
```diff
@@ -23,34 +23,48 @@ resource "pagerduty_ruleset" "foo" {
...
+      # Until 11:30am (2 hours later)
+      duration = "${2 * 60 * 60 * 1000}"
```
stmcallister, line 46: Similar as the above comment. Please use the v0.12+ syntax for variables.

website/docs/r/ruleset_rule.html.markdown
```diff
@@ -109,34 +123,34 @@ The following arguments are supported:
...
+  * `start_time` - A Unix timestamp in milliseconds which is combined with the `timezone` to determine the time this rule will start on each specified `weekday`. Note that the _date_ of the timestamp you specify does **not** matter, except that it lets you determine whether or not daylight saving time is in effect so that you use the correct UTC offset for the timezone you specify. In practice, you may want to use [the `time_static` resource](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/static) to generate this value, as demonstrated in the `resource.pagerduty_ruleset_rule.foo` code example at the top of this page. To generate this timestamp manually, if you want your rule to apply starting at 9:30am in the `America/New_York` timezone, use your programing language of choice to determine a Unix timestamp that represents 9:30am in that timezone, like [1554989400000](https://www.epochconverter.com/timezones?q=1554989400000&tz=America%2FNew_York).
+  * `duration` - Length of time the schedule will be active in milliseconds. For example `duration = "${2 * 60 * 60 * 1000}"` if you want your rule to apply for 2 hours, from the specified `start_time`.
```
stmcallister, line 150: Also here, on the comments from above re: v0.12+ syntax.

stmcallister (changes requested):
Overall, looks great! Please just update the variables to use the current syntax. Thanks!

metavida:
Ah! Good catch, @stmcallister ! I've removed those 3 cases of "interpolation-only expressions" in e3f0ec6f1dc89847061d18c7e874e77b90cd7dc3 and tested locally to confirm that the syntax still works as expected 👍

stmcallister (approved):
👍 🎉  Awesome! Thank you! LGTM.

metavida:
Thanks so much for getting this merged! Will the documentation changes to live at https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/ruleset_rule when we (i.e. you) release a new version of the terraform provider?

Issue #200, opened 2020-04-09
spurious diff on pagerduty_schedule


rblumen-desk:
### Terraform Version
v0.12.24
### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_schedule

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
terraform {
  required_version = ">= 0.12.20"

  backend local {
    path = "tfstate"
  }
}

provider pagerduty {}
variable team_members {
  type        = list(string)
  description = "list of emails of the team members in the on call rotations"
  default     = ["<redacted>"]
}

variable time_zone {
  description = "time zone in which the schedule is defined.  the value must be a string from the [tz database](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)"
  default     = "America/Los_Angeles"
}

variable rotation_start_time {
  description = "all on-call rotations will start at ths time (in RFC3339 format)"
  default     = "2019-12-03T00:00:00.00-08:00"
}
data pagerduty_user team_members {
  for_each = toset(var.team_members)
  email    = each.value
}

resource pagerduty_team team {
  name        = "BUG team"
  description = "BUG team"
}

resource pagerduty_team_membership membership {
  for_each = data.pagerduty_user.team_members
  team_id  = pagerduty_team.team.id
  user_id  = each.value.id
}

resource pagerduty_schedule on_call {
  name        = "BUG on call"
  time_zone   = var.time_zone
  description = "BUG on call"

  layer {
    name                         = "layer 1"
    start                        = var.rotation_start_time
    rotation_virtual_start       = var.rotation_start_time
    rotation_turn_length_seconds = 604800
    users                        = values(data.pagerduty_user.team_members)[*].id
  }
}
```

### Debug Output
Not available.

### Panic Output
N/A

### Expected Behavior
The plan should not show any changes

### Actual Behavior
Spurious diffs were shown
```
Terraform will perform the following actions:

  # pagerduty_schedule.on_call will be updated in-place
  ~ resource "pagerduty_schedule" "on_call" {
        description = "BUG on call"
        id          = "<redacted>"
        name        = "BUG on call"
        time_zone   = "America/Los_Angeles"

      ~ layer {
            id                           = "<redacted>"
            name                         = "layer 1"
            rotation_turn_length_seconds = 604800
[...]

PR #343, opened 2021-06-08
resource/pagerduty_schedule: suppress spurious start diff

Padorax:
Issue: https://github.com/PagerDuty/terraform-provider-pagerduty/issues/200

When both schedule layer start time (the old one read from PagerDuty API, and the new one read
from configuration files) are in the past, there is no need to show Diff as the acutual start
time will be updated by PagerDuty API to current time.

Otherwise, users will have to sync the configuration files from time to time, when they make
irrelevant changes to their schedules (e.g., adding a user) or even no changes at all.


stmcallister (approved):
Thanks for this! Looks great! 🎉 👍

Issue #345, opened 2021-06-11
`pagerduty_team` should return a list of members


TomTucka:
Heya,

The `pagerduty_team` data source should be able to return a list of team members, I've added an example use case below! 

### Affected Resource(s)
- pagerduty_team (d)

### Terraform Configuration Files
```hcl
data "pagerduty_team" "my_team" {
  name = "My Team Name"
}

resource "pagerduty_schedule" "my_schedule" {
  name      = "My Schedule"
  time_zone = "Europe/London"

  layer {
    name                         = "Layer 1"
    rotation_turn_length_seconds = 604800
    rotation_virtual_start       = "2019-08-19T09:00:00+01:00"
    start                        = "2019-07-26T12:15:46+01:00"
    users                        = data.pagerduty_team.my_team.members
  }
}


```
### Expected Behavior
`pagerduty_team` should return a list of members

### References
- https://github.com/heimweh/go-pagerduty/blob/master/pagerduty/team.go#L146

imjaroiswebdev:
Hi @TomTucka to solve this request We have added the new `users` **data source** for querying the list of users from the PagerDuty account, which allows to filter by one or more teams at the same time. This new **data source** will be available with our next release of the provider.

PR #346, opened 2021-06-11
Return a list of users for `pagerduty_team`

TomTucka:
Still WIP, Need to figure out tests for this 👍🏼

### DIFF
--- a/pagerduty/data_source_pagerduty_team.go
+++ b/pagerduty/data_source_pagerduty_team.go
@@ -28,6 +28,18 @@ func dataSourcePagerDutyTeam() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"members": {
+				Type:     schema.TypeList,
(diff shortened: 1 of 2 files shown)


stmcallister:
I like where you're going with this! Once you get some tests figured out and implemented I'd be happy to review!

jbfavre:
I second this PR, because it could help to manage user_notifications using Terraform, without having the need to manage users
It could be the opportunity to enforce users' notifications rules at team level.

However, I also see a possible issue with the logic exposed in #345 : when creating a new layer using the Web UI, and depending on start date, users' will be sorted out by PagerDuty API to keep the schedule order consistent.
So, how to sort team members list to be usable in a layer configuration ?
I don't see any elegant way to automatically reconciliate Terraform and the API. :thinking: 
WDYT @stmcallister ?

doryer:
Looking forward for this to be merged

PR #348, opened 2021-06-23
Add ServiceNow extension resource

drastawi:
Adding a dedicated resource for ServiceNow.

This resource will allow the end-user to avoid perma-diff and mark the password as a sensitive value while providing a simpler dedicated interface (e.g. endpoint URL can be very easily mistaken for the ServiceNow target).

### DIFF
--- a/pagerduty/resource_pagerduty_extension_servicenow_test.go
+++ b/pagerduty/resource_pagerduty_extension_servicenow_test.go
@@ -1,0 +1,215 @@
+package pagerduty
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func init() {
+	resource.AddTestSweepers("pagerduty_extension_servicenow", &resource.Sweeper{
+		Name: "pagerduty_extension_servicenow",
+		F:    testSweepExtensionServicenow,
+	})
+}
+
+func testSweepExtensionServicenow(region string) error {
+	config, err := sharedConfigForRegion(region)
+	if err != nil {
+		return err
+	}
+
+	client, err := config.Client()
+	if err != nil {
+		return err
+	}
+
+	resp, _, err := client.Extensions.List(&pagerduty.ListExtensionsOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, extension := range resp.Extensions {
+		if strings.HasPrefix(extension.Name, "test") || strings.HasPrefix(extension.Name, "tf-") {
+			log.Printf("Destroying extension %s (%s)", extension.Name, extension.ID)
+			if _, err := client.Extensions.Delete(extension.ID); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func TestAccPagerDutyExtensionServicenow_Basic(t *testing.T) {
+	extension_name := resource.PrefixedUniqueId("tf-")
+	extension_name_updated := resource.PrefixedUniqueId("tf-")
+	name := resource.PrefixedUniqueId("tf-")
+	url := "https://example.com/recieve_a_pagerduty_webhook"
+	url_updated := "https://example.com/webhook_foo"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckPagerDutyExtensionServicenowDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckPagerDutyExtensionServicenowConfig(name, extension_name, url, "false", "any"),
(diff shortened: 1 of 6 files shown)


pagerduty/resource_pagerduty_extension_servicenow_test.go
```diff
@@ -0,0 +1,193 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckPagerDutyExtensionServicenowExists("pagerduty_extension_servicenow.foo"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "name", extension_name),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "extension_schema", "PJFWPEP"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "endpoint_url", url),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "config", "{\"notify_types\":{\"acknowledge\":false,\"assignments\":false,\"resolve\":false},\"restrict\":\"any\"}"),
```
stmcallister, line 72: There is no longer a `config` attribute in this resource. Please break out these checks to match the current schema of the resource.

pagerduty/resource_pagerduty_extension_servicenow_test.go
```diff
@@ -0,0 +1,193 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckPagerDutyExtensionServicenowExists("pagerduty_extension_servicenow.foo"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "name", extension_name_updated),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "extension_schema", "PJFWPEP"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "endpoint_url", url_updated),
+					resource.TestCheckResourceAttr(
+						"pagerduty_extension_servicenow.foo", "config", "{\"notify_types\":{\"acknowledge\":true,\"assignments\":true,\"resolve\":true},\"restrict\":\"pd-users\"}"),
```
stmcallister, line 88: Same as above. Please break out these checks to match the current schema

pagerduty/resource_pagerduty_extension_servicenow_test.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+data "pagerduty_extension_schema" "foo" {
+	name = "Generic V2 Webhook"
+}
+
+resource "pagerduty_extension_servicenow" "foo"{
+  name = "%s"
+  endpoint_url = "%s"
+  extension_schema = data.pagerduty_extension_servicenow_schema.foo.id
```
stmcallister, line 182: I think this was supposed to be referencing `data.pagerduty_extension_schema.foo.id`

pagerduty/resource_pagerduty_extension_servicenow_test.go
```diff
@@ -0,0 +1,193 @@
...
+  extension_objects = [pagerduty_service.foo.id]
+  snow_user = "meeps"
+  snow_password = "zorz"
+  sync_options = "manual_sync"
+  target = foo.servicenow.com/webhook_foo
```
stmcallister, line 187: I think the `target` value is supposed to be a string.

website/docs/r/extension_servicenow.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+```hcl
+data "pagerduty_extension_schema" "webhook" {
+  name = "Generic V2 Webhook"
+}
+
+resource "pagerduty_user" "example" {
+  name  = "Howard James"
+  email = "howard.james@example.domain"
+  teams = [pagerduty_team.example.id]
```
stmcallister, line 23: The `teams` field is actually deprecated in the `pagerduty_user` resource. For these docs you can just remove the field.

website/docs/r/extension_servicenow.html.markdown
```diff
@@ -0,0 +1,87 @@
...
+      id   = pagerduty_user.example.id
+    }
+  }
+}
+
+resource "pagerduty_service" "example" {
+  name                    = "My Web App"
+  auto_resolve_timeout    = 14400
+  acknowledgement_timeout = 600
+  escalation_policy       = pagerduty_escalation_policy.example.id
```
stmcallister, line 44: Since we're making other edits to this doc. This should probably match the definition above for `pagerduty_escalation_policy.foo.id`.

stmcallister (changes requested):
Thanks for submitting this! I like the direction of this PR. I think it's probably cleaner for extensions like ServiceNow to have their own resource. However, there are several errors in the tests. I've pointed out a few that I see. If you could fix those I'd be happy to merge.

drastawi:
Hey @stmcallister, I ran all the tests now against a live environment. Should all work. Also fixed the corresponding typos on the non-servicenow extension side. Let me know if it looks better now.

stmcallister (approved):
Thank you! Also, I think `Servicenow` should be `ServiceNow` throughout the code. But, i can make that change before merging. Appreciate the contribution!! 🎉

PR #349, opened 2021-06-23
Update ruleset_rule.html.markdown

pdt-dridge:
Updating the route value to use a Terraform resource rather than a hard coded Service ID.


stmcallister (approved):
Thanks, Ridgey! 🎉

PR #350, opened 2021-06-24
Set MinItems: 1 on schedule layers

drastawi:
Empty users returns a 400 on the PD API.


stmcallister (approved):
Nice improvement! 🎉 👍 Thanks!

PR #363, opened 2021-07-29
add integration datasource

petetanton:
This PR adds support for `data.pagerduty_service_integration` so that integration keys can be fetched from PagerDuty when configuring alerts from third party systems in Terraform

Example usage:
```hcl
data "pagerduty_integration" pdint {
  service_name = "my-service-name"
  integration_summary = "Amazon CloudWatch"
}
```

### DIFF
--- a/pagerduty/data_source_pagerduty_service_integration_test.go
+++ b/pagerduty/data_source_pagerduty_service_integration_test.go
@@ -2,54 +2,12 @@ package pagerduty
 
 import (
 	"fmt"
-	"os"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
-	"github.com/heimweh/go-pagerduty/pagerduty"
 )
-
-func setupResources(service, serviceIntegration, escalationPolicy, username, email string) error {
-	client, err := pagerduty.NewClient(&pagerduty.Config{Token: os.Getenv("PAGERDUTY_TOKEN")})
-	if err != nil {
-		return err
-	}
-
-	createUser, _, err := client.Users.Create(&pagerduty.User{Name: username, Email: email})
-	if err != nil {
-		return err
-	}
-
-	createEp, _, err := client.EscalationPolicies.Create(&pagerduty.EscalationPolicy{
-		Name:            escalationPolicy,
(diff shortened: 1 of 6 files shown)


petetanton:
@stmcallister please can I get a review on this? I am not sure exactly what the process is but more than happy add any clarification if required!

gsreynolds (approved):
Looks great! Thanks for adding the documentation too 🎉 

Test results:

```
TF_ACC=1 go test -run "TestAccDataSourcePagerDutyIntegration_Basic" ./pagerduty -v -timeout 120m
=== RUN   TestAccDataSourcePagerDutyIntegration_Basic
--- PASS: TestAccDataSourcePagerDutyIntegration_Basic (8.28s)
PASS
ok      github.com/terraform-providers/terraform-provider-pagerduty/pagerduty   9.008s
```

pagerduty/data_source_pagerduty_service_integration_test.go
```diff
@@ -0,0 +1,105 @@
...
+		Service: &pagerduty.ServiceReference{ID: createResponse.ID, Type: "service_reference"},
+		Vendor:  &pagerduty.VendorReference{ID: "PAM4FGS", Type: "vendor_reference"},
+	})
+	if err != nil {
+		return fmt.Errorf("error creating integration: %v", err)
+	}
+
+	return nil
+}
+
```
stmcallister, line 53: When I first saw this I thought, "wow! what a super clean way to do this!" But after running the tests a few times I discovered why we don't do it this way through out the rest of the provider. When you create the objects directly through the client then the Terraform test suite doesn't manage them. So, when the tests are done all these created objects still exist on the PagerDuty account. When you define these objects in the HCL below then Terraform Testing will manage them correctly (ie, create the objects and then delete them when the test is complete. 

Please move the creation of these dependent objects to the HCL below. (Good example is in the [data_source_pagerduty_service_test](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/273f040555279cb2f9e4d06b422add3087440e0f/pagerduty/data_source_pagerduty_service_test.go#L57))

stmcallister (changes requested):

pagerduty/data_source_pagerduty_service_integration_test.go
petetanton: Thanks for the feedback! I have tried to implement your suggestion.
I ended up doing a test within a test so that resources created in the first plan can be reference by the second plan.
I am new to terraform testing so please feel free to suggest alternatives! 

Thanks
Pete

stmcallister (approved):
It is a different approach than the other tests, but from what I can tell it seems to do the job. Thanks for adding this feature to the provider!! 👍 🎉

PR #364, opened 2021-08-01
Add type as computed attributes to service & business_service data sources + resources

jjm:
As was discussed on this weeks PagerDuty Terraform round table, raising a PR to add a `dependency_type` commuted attribute for the following:

- `data pagerduty_service`
- `data pagerduty_business_service`
- `resource pagerduty_service`
- `pagerduty_business_service`

Also a correct the repo name in the README, as the repo is now owned by PagerDuty.

**ToDo:**
- [x] Update docs
- [x] Update tests.

### DIFF
--- a/pagerduty/resource_pagerduty_business_service_test.go
+++ b/pagerduty/resource_pagerduty_business_service_test.go
@@ -71,6 +71,8 @@ func TestAccPagerDutyBusinessService_Basic(t *testing.T) {
 						"pagerduty_business_service.foo", "point_of_contact", pointOfContact),
 					resource.TestCheckResourceAttrSet(
 						"pagerduty_business_service.foo", "self"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_business_service.foo", "dependency_type", pointOfContact),
 				),
 			},
 			{
--- a/pagerduty/resource_pagerduty_business_service_test.go
+++ b/pagerduty/resource_pagerduty_business_service_test.go
@@ -72,7 +72,7 @@ func TestAccPagerDutyBusinessService_Basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(
 						"pagerduty_business_service.foo", "self"),
 					resource.TestCheckResourceAttr(
-						"pagerduty_business_service.foo", "dependency_type", pointOfContact),
+						"pagerduty_business_service.foo", "dependency_type", "business_service"),
 				),
 			},
 			{
--- a/pagerduty/resource_pagerduty_business_service_test.go
+++ b/pagerduty/resource_pagerduty_business_service_test.go
@@ -72,7 +72,7 @@ func TestAccPagerDutyBusinessService_Basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(
 						"pagerduty_business_service.foo", "self"),
 					resource.TestCheckResourceAttr(
-						"pagerduty_business_service.foo", "dependency_type", "business_service"),
+						"pagerduty_business_service.foo", "type", "business_service"),
 				),
 			},
 			{
--- a/website/docs/d/business_service.html.markdown
+++ b/website/docs/d/business_service.html.markdown
@@ -27,5 +27,6 @@ The following arguments are supported:
 ## Attributes Reference
 * `id` - The ID of the found business service.
 * `name` - The short name of the found business service.
+* `dependency_type` - The type for passing to a service dependency.
 
 [1]: https://api-reference.pagerduty.com/#!/Business_Services/get_business_services
--- a/website/docs/d/business_service.html.markdown
+++ b/website/docs/d/business_service.html.markdown
@@ -27,6 +27,6 @@ The following arguments are supported:
 ## Attributes Reference
 * `id` - The ID of the found business service.
 * `name` - The short name of the found business service.
-* `type` - The type for passing to a service dependency.
+* `type` - The type of object. The value returned will be `business_service`. Can be used for passing to a service dependency.
 
-[1]: https://api-reference.pagerduty.com/#!/Business_Services/get_business_services
+* [1]: https://api-reference.pagerduty.com/#!/Business_Services/get_business_services
--- a/website/docs/d/business_service.html.markdown
+++ b/website/docs/d/business_service.html.markdown
@@ -27,6 +27,6 @@ The following arguments are supported:
 ## Attributes Reference
 * `id` - The ID of the found business service.
 * `name` - The short name of the found business service.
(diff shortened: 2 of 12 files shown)


stmcallister:
@jjm Super cool idea, as we discussed in the roundtable meeting. Remind me how the fact that the Business Service already having a `type` field is a blocker? 🤔

jjm:
@stmcallister It was the following `type` in the schema for `resourcePagerDutyBusinessService` that I hit a block:

https://github.com/jjm/terraform-provider-pagerduty/blob/170b467d47a2e673a57ded62ca43687c5edb424d/pagerduty/resource_pagerduty_business_service.go#L43-L50

As I had added a `type` to the 2 data sources and to `resourcePagerDutyService`.

I think maybe the best option is to add a `dependancy_type` to these resources & data sources instead, so that they don't have a clash with the existing schema.

jjm:
I've successfully tested this change locally built, with the following config:

```terraform
terraform {
  required_providers {
    pagerduty = {
      source  = "terraform.example.com/pagerduty/pagerduty"
      version = "2.3.0"
    }
  }
}

provider "pagerduty" {}

data "pagerduty_escalation_policy" "default" {
  name = "Default"
}

resource "pagerduty_business_service" "checkout_app" {
  name = "Checkout App"
}

resource "pagerduty_service" "web_app" {
  name                    = "Our Web App"
  auto_resolve_timeout    = 14400
  acknowledgement_timeout = 600

  escalation_policy = data.pagerduty_escalation_policy.default.id
}

resource "pagerduty_service" "shopping_cart" {
  name                    = "Shopping Cart"
  auto_resolve_timeout    = 14400
  acknowledgement_timeout = 600

  escalation_policy = data.pagerduty_escalation_policy.default.id
}

resource "pagerduty_service" "vault" {
  name                    = "Vault"
  auto_resolve_timeout    = 14400
  acknowledgement_timeout = 600

  escalation_policy = data.pagerduty_escalation_policy.default.id
}

locals {
  vault = [
    pagerduty_business_service.checkout_app,
    pagerduty_service.shopping_cart,
    pagerduty_service.web_app
  ]
}

resource "pagerduty_service_dependency" "vault" {
  for_each = { for k in local.vault : k.id => k.dependency_type }

  dependency {
    dependent_service {
      id   = each.key
      type = each.value
    }
    supporting_service {
      id   = pagerduty_service.vault.id
      type = pagerduty_service.vault.dependency_type
    }
  }
}
```

The `for_each` using the local was just what I was wanting to support. The services need created before the dependancies in this example due to Terraform not being able to determine the number of resources. However other use cases will be possible.

jjm:
@stmcallister Ready for review now and the workflow needs to be approved as this is my first contribution.

pagerduty/resource_pagerduty_business_service_test.go
```diff
@@ -71,6 +71,8 @@ func TestAccPagerDutyBusinessService_Basic(t *testing.T) {
 						"pagerduty_business_service.foo", "point_of_contact", pointOfContact),
 					resource.TestCheckResourceAttrSet(
 						"pagerduty_business_service.foo", "self"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_business_service.foo", "dependency_type", pointOfContact),
```
jjm, line 75: I think this should have been:

```suggestion
						"pagerduty_business_service.foo", "dependency_type", "business_service"),
```

stmcallister:
Thanks @jjm! I love this idea in concept, but still not a fan of adding an additional field to the resources that doesn't exist on the PagerDuty API, if we don't have to. 

On the `pagerduty_business_service` resource there is already a `type` field, as you mentioned, that is constrained to `business_service` or `business_service_reference`. Making this field editable in the provider was my mistake from the get go because the API ignores it on `POST` and `PUT`. And, if someone sets the value to `business_service_reference` in the provider they'll get stuck a permanent diff.  However, there still is a possibility that some users are setting the value in their scripts.

With all of that said, I think we should change `dependency_type` to just `type` on the `pagerduty_service` resource.

For the `pagerduty_business_service` resource we have a couple of options I would love your input on.  I would like to use the existing `type` field. To make sure it's going to provide a value that can be used we can either:

1. Make the `type` field computed, thus introducing a breaking change, but staying uniform with the API.
2. Leave the field "editable" to not break existing scripts but change the validation to only accept `business_service` as a value.

What are your thoughts as a user?

jjm:
@stmcallister Thanks, I think `type` over `dependency_type` does makes sense.

I think going with the second option is best, but also mark this attribute as **Deprecated** following https://www.terraform.io/plugin/sdkv2/best-practices/deprecations#provider-attribute-removal and when there's a major version bump the `Deprecated` flag is changed to `Commuted`.

I'll make these changes later on today. I think causing a validation failure when there's a constant diff happening is not so much of an issue.

jjm:
@stmcallister I've made those changes.

website/docs/d/business_service.html.markdown
```diff
@@ -27,5 +27,6 @@ The following arguments are supported:
 ## Attributes Reference
 * `id` - The ID of the found business service.
 * `name` - The short name of the found business service.
+* `type` - The type for passing to a service dependency.
```
stmcallister, line 30: I would say something like

```
`type` - The type of object. The value returned will be `business_service`. Can be used for passing to a service dependency.

File: website/docs/d/service.html.markdown [hunk 3]
```diff
@@ -39,5 +39,6 @@ The following arguments are supported:
 
 * `id` - The ID of the found service.
 * `name` - The short name of the found service.
+* `type` - The type for passing to a service dependency.
```
stmcallister, line 42: Same as above. Technically, the types are for a variety of things and not just setting service dependency.

website/docs/r/service.html.markdown
```diff
@@ -154,6 +154,7 @@ The following attributes are exported:
   * `created_at`- Creation timestamp of the service.
   * `status`- The status of the service.
   * `html_url`- URL at which the entity is uniquely displayed in the Web app.
+  * `type` - The type for passing to a service dependency.
```
stmcallister, line 157: Also, same as above.

stmcallister (changes requested):
Looks great! Just a few tweaks to the docs and we'll be good. Thanks @jjm !!

website/docs/d/business_service.html.markdown
jjm: Thanks, I just used that.

jjm:
Thanks @stmcallister, I've addressed your review comments.

stmcallister (approved):
🎉 🌮 👍 Awesome! Thank you @jjm !!!

PR #366, opened 2021-08-04
Add optional base_url configuration

busatol:
This pr adds the optional `base_url` configuration argument that is passed to the API client. 

```terraform
provider "pagerduty" {
  token     = "xxxxxxxxxxxx"
  base_url  = "https://custom-cache-proxy.com"
}
```

### DIFF
--- a/pagerduty/config.go
+++ b/pagerduty/config.go
@@ -14,6 +14,9 @@ type Config struct {
 	// The PagerDuty API URL
 	ApiUrl string
 
+	// Override default PagerDuty API URL
+	ApiUrlOverride string
+
 	// The PagerDuty APP URL
 	AppUrl string
 
@@ -48,8 +51,13 @@ func (c *Config) Client() (*pagerduty.Client, error) {
 	httpClient = http.DefaultClient
 	httpClient.Transport = logging.NewTransport("PagerDuty", http.DefaultTransport)
 
+	var apiUrl = c.ApiUrl
+	if c.ApiUrlOverride != "" {
+		apiUrl = c.ApiUrlOverride
+	}
+
 	config := &pagerduty.Config{
-		BaseURL:    c.ApiUrl,
+		BaseURL:    apiUrl,
 		Debug:      logging.IsDebugOrHigher(),
 		HTTPClient: httpClient,
 		Token:      c.Token,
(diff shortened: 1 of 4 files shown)


stmcallister:
Thank you for submitting this! However, I think I want to go a slightly different direction. Rather than having folks set the URLs directly I think we should just provide a `service_region` field to the provider tag. Then the provider will figure out which URLs to use. At least that's my thoughts right now.

From a user perspective, do you think that would be a cleaner approach?

rymir:
Hi @stmcallister, thanks for looking into this.

What we're trying to do here is actually setting a totally custom proxy endpoint, running internally in our infrastructure. We're using these custom endpoints to cache http GETs therefore reducing the throughput to the 3rd party providers (pagerduty in this case) and reduce the chances of hitting 3rd party providers rate limiting throttling.

Of course, as you mention it can be used to force traffic to a specific pagerduty "service_region" endpoint, but we need to be able to set "any" configurable endpoint instead of just a subset of them managed by the provider itself.

rymir:
@stmcallister any updates?

stmcallister:
Ah, I hadn't thought of that scenario. 🤔 Let me think about this one and talk to a few folks.

busatol:
> @busatol It does. I feel the `service_region` logic is going to be more useful to more users. Also, with the addition of the `slack_connection` resource we now need to support multiple baseURLs. All the other resources use `api.pagerduty.com` while the `slack_connection` resource uses the `app.pagerduty.com` url. However, if your use case, you would only need one `base_url` because it's going to be the address of your proxy. Right?
> 
> How about we name your field something like `override_base_url` and then mention in the docs and README that it's used for proxy use-cases? That way other users won't get confused about whether to use the field for setting EU urls, etc.

Updated. I added the field `api_url_override`.

stmcallister:
@busatol Thanks for making the naming change. Looks like this branch is still conflicting with master. Will you merge with the latest master and fix the conflicts? Once I can test it successfully i'll approve and merge.

busatol:
> @busatol Thanks for making the naming change. Looks like this branch is still conflicting with master. Will you merge with the latest master and fix the conflicts? Once I can test it successfully i'll approve and merge.

@stmcallister done. The PR is up to date with master 👍

stmcallister (approved):
Looks good! thanks for fixing the merge conflicts! 🎉

Issue #370, opened 2021-08-13
Password set to blank on change to pagerduty_extension_servicenow


martindstone:
Terraform sets snow_password to blank when ignore_changes is set on snow_password and any other extension config element is changed. I will be opening a PR with a fix shortly.

### Terraform Version
1.10.0

### Affected Resource(s)
pagerduty_extension_servicenow

### Terraform Configuration Files
```hcl
data "pagerduty_extension_schema" "webhook" {
  name = "ServiceNow (v7)"
}

resource "pagerduty_user" "example" {
  name  = "TF SNOW Test User"
  email = "martin+snowtf@pagerduty.com"
}

resource "pagerduty_escalation_policy" "example" {
  name      = "TF SNOW Test Escalation Policy"
  num_loops = 2

  rule {
    escalation_delay_in_minutes = 10

    target {
      type = "user"
      id   = pagerduty_user.example.id
    }
  }
}

resource "pagerduty_service" "example" {
  name                    = "TF SNOW Test Service"
  auto_resolve_timeout    = 14400
  acknowledgement_timeout = 600
  escalation_policy       = pagerduty_escalation_policy.example.id
}


resource "pagerduty_extension_servicenow" "snow"{
  name = "TF SNOW Test Extension"
  extension_schema = data.pagerduty_extension_schema.webhook.id
  extension_objects = [pagerduty_service.example.id]
  snow_user = "meeps"
  snow_password = "zorz"
  sync_options = "manual_sync"
  target = "https://foo.servicenow.com/webhook_foo"
  task_type = "incident"
  referer = "None"
  lifecycle {
    ignore_changes = [
      snow_password,
    ]
  }
}
```

and then

```hcl
data "pagerduty_extension_schema" "webhook" {
  name = "ServiceNow (v7)"
}

resource "pagerduty_user" "example" {
  name  = "TF SNOW Test User"
  email = "martin+snowtf@pagerduty.com"
}

resource "pagerduty_escalation_policy" "example" {
  name      = "TF SNOW Test Escalation Policy"
  num_loops = 2

  rule {
    escalation_delay_in_minutes = 10

    target {
      type = "user"
      id   = pagerduty_user.example.id
    }
  }
}

resource "pagerduty_service" "example" {
  name                    = "TF SNOW Test Service"
  auto_resolve_timeout    = 14400
  acknowledgement_timeout = 600
  escalation_policy       = pagerduty_escalation_policy.example.id
}


[...]

PR #371, opened 2021-08-13
don't send blank snow_password when changes are ignored

martindstone:
This is a fix for issue #370. snow_password should not be sent when it's blank (due to lifecycle ignore_changes)


stmcallister (approved):
👍 Looks good!

PR #379, opened 2021-08-27
user: Validate role against known role names

zecke:


pagerduty/resource_pagerduty_user.go
```diff
@@ -44,6 +44,17 @@ func resourcePagerDutyUser() *schema.Resource {
...
+				ValidateFunc: validateValueFunc([]string{
+					"admin",
+					"limited_user",
+					"observer",
+					"owner",
+					"read_only_user",
+					"restricted_access",
+					"read_only_limited_user",
+					"user",
+					"team_responder",
```
drastawi, line 56: Looks like the 'team_responder' role has been deprecated: https://community.pagerduty.com/forum/t/creating-user-via-api-with-role-team-responder-sets-observer/1794
https://gist.github.com/fussybeaver/4ba9a98fed7c0c9c836a04ade3a108f3#file-json-L1836

Looks good otherwise.

zecke:
@stmcallister PTAL

stmcallister (changes requested):
Thanks for this! @drastawi is correct. Can you remove `team_responder`? Also, the `owner` role while a valid role can't be set through the API either, so please remove that one as well. Account owners actually [can't even be updated through the API at all](https://community.pagerduty.com/forum/t/unable-to-edit-account-owner-via-api-or-terraform/2297).

zecke:
Sure. I can remove `team_responder` but `owner` is tricky. We do have this field set in our config AFAICT the terraform apply succeeds (because the owners are already owners?). What role should we put for users that are already owners?

stmcallister:
@zecke It was my understanding that you [could not manage Account Owners through the API](https://community.pagerduty.com/forum/t/unable-to-edit-account-owner-via-api-or-terraform/2297). But you're seeing this in your Terraform configs?

drastawi:
@stmcallister I am listed as an account owner of my org and I do have myself listed in terraform with an owner role. 

It is true the 403 happens when I try to change my role, but it is still useful to have myself listed under terraform as I can more easily reference my ID to be put on schedules/escalation policies, get associated with contact methods. I am listed as the owner on a publicly viewable repo, so others can easily see who to reach out to when they need to make changes that only an account owner has rights to make.

When we want to change an owner, we make the change in the code and commit it, but then change the ownership manually.

pagerduty/resource_pagerduty_user.go
zecke: removed it.

Issue #365, opened 2021-08-02
Slack v2 next generation support


jbfavre:
Hi there,

Now that Slack v2 next generation is in early access (https://support.pagerduty.com/docs/slack-v2-next-generation), I wonder when it'll be supported by Terraform provider.
As far as I understood, it'll make Terraform much easier to manage: the authentication is only need once at workspace level.
Connexions from team or services to Slack channel will reuse it which could allow to fully declare it through Terraform.

Do you have any plan to support Slack v2 next generation integration ?

stmcallister:
I believe #381 addressed this issue. Going to close. Please reopen if it's not the case.

Issue #357, opened 2021-07-14
Slack V2 Extension Schema No Longer Found


christopherbox:
### Terraform Version
1.0.2

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_extension_schema

### Terraform Configuration Files
```hcl
data "pagerduty_extension_schema" "slack" {
  name = "Slack V2"
}
```

### Expected Behavior
Not even sure, I understand that it got removed from the Extension Schema's from the API. I would be okay if I was able to move it somewhere else or something.

### Actual Behavior
Get an error that the extension schema with the name Slack V2 cannot be found
```hcl
Error: Unable to locate any extension schema with the name: Slack V2
```

### Steps to Reproduce
1. Configure a data resource
2. `terraform init`
3. `terraform plan`

### Important Factoids
The API call for extension_schemas no longer returns an entry for Slack at all. I see that there is now a Slack Integration API. My assumption is that a new slack resource might be made to handle this new logic? However between now and then, all of our PagerDuty plans fail because of this all of a sudden dissapearance with the extension schema. In order to continue, I am going to have to either target specific applies or delete the entry in the state file and remove the code.

jbfavre:
:wave: 
Same issue here.
I think it's worth mentionning it happened right after I enable the `Slack v2 new generation` integration.
What is the new schema name ?
Any update to do on PagerDuty terraform provider ?

jbfavre:
For what's worth, I found a quick & dirty fix to workaround the issue:
```
data "pagerduty_extension_schema" "slack" {
  name = "Slack V2"
}
```
doesn't exist anymore on PagerDuty side, but it's still referenced into our Terraform state.
What I did is to grab the ID from Terraform state and hard code it into all existing slack integration resources.
Now, existing Slack integration resources are fine and our Terraform workflow is back.

However, I've no idea on wether it'll work for new slack integration resources.

Hope this helps

jbfavre:
Looks like the Slack v2 extension schema is back, at least for me
Did nothing except opening a case to PagerDuty support :man_shrugging: 
No explanation so far of what happened

sheenamt:
I'm having this same issue, using terraform 1.0.1

jrsdav:
> Looks like the Slack v2 extension schema is back, at least for me
> Did nothing except opening a case to PagerDuty support 🤷‍♂️
> No explanation so far of what happened

[...]

Issue #280, opened 2020-11-11
Permadiff on Slack Integration


sparky005:
Terraform shows a permadiff on some slack integrations with some services. This only happens on a handful of services in our environment. I believe there are some extra keys in the json config for those services that TF is trying to clear out and that Pagerduty then re-creates. 

### Terraform Version
```
Terraform v0.13.4
+ provider registry.terraform.io/hashicorp/aws v3.13.0
+ provider registry.terraform.io/terraform-providers/datadog v2.15.0
+ provider registry.terraform.io/terraform-providers/pagerduty v1.7.11
+ provider terraform.better.com/yougroupteam/buildkite v0.0.9
```

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_extension.slack

### Terraform Configuration Files
```hcl
# NB: vars are set elsewhere, I'm not including them as I don't think they're relevant to this issue
# find slack v2 extension schema
data "pagerduty_extension_schema" "slack" {
  name = "Slack V2"
}

# configure extension
resource "pagerduty_extension" "slack" {
  name              = title(var.service_name)
  extension_schema  = data.pagerduty_extension_schema.slack.id
  extension_objects = ["${pagerduty_service.default.id}"]

  config = <<EOF
{
  "notify_types": {
    "resolve": true,
    "trigger": true,
    "escalate": true,
    "acknowledge": true,
    "assignments": true,
    "annotate": true
  },
  "urgency": {
    "high": true,
    "low": true
  },
  "referer": "${pagerduty_service.default.html_url}/integrations",
  "ok": true,
  "scope": "identify,bot,commands,incoming-webhook,channels:read,groups:read,im:read,team:read,users:read,users:read.email,channels:write,chat:write:user,chat:write:bot,groups:write",
  "user_id": "${var.slack_user_id}",
  "team_id": "${var.slack_team_id}",
  "enterprise_id": "${var.slack_enterprise_id}",
  "team_name": "${var.slack_team_name}",
  "incoming_webhook": {
    "channel": "${var.slack_channel_name}",
    "channel_id": "${var.slack_channel_id}",
    "configuration_url": "${var.slack_configuration_url}",
    "url": "${var.slack_webhook_url}"
  },
  "bot": {
    "bot_user_id": "${var.slack_bot_user_id}"
  },
  "access_token": "redacted"
}
EOF
}

```

### Debug Output
Not including for now as it contains tokens, I'll upload if it's needed

[...]

PR #381, opened 2021-08-31
Add Slack v2 Connection Resource

### LINKED ISSUE #73: unable to use terraform-provider-pagerduty to create a Slack extension
$ terraform -v
Terraform v0.10.8

aws_provider_version = "1.9"
pagerduty_provider_version = "1.0"

Using the brand new pagerduty provider extension (thanks much for that!)
https://github.com/terraform-providers/terraform-provider-pagerduty/pull/69

```
data "pagerduty_extension_schema" "webhook" {
  name = "Slack"
}

resource "pagerduty_extension" "slack" {
  name = "name"
  endpoint_url = "https://hooks.slack.com/services/redacted"
  extension_schema = "${data.pagerduty_extension_schema.webhook.id}"
  extension_objects    = ["${pagerduty_service.service.id}"]
}
```

If I use the "Slack" pagerduty_extension_schema, Terraform creates a new extension.  However the extension doesn't work (send messages to the appropriate slack channel) unless I authorize it in the Pagerduty console , and the icon in Pagerduty under Configuration / Extensions is broken (circle w/ a slash through it).  If I then edit the new extension in Pagerduty, select the Slack channel, and authorize the extension, it will then work.

Questions - 
1. how do I fully automate creating an extension?  I don't see a way to authorize the extension and tell Pagerduty which slack channel to use for the notification using new extension
2. is there a way to use "Generic V2 Webhook" as a Slack extension?  I'm just looking for a way to make this work easily, I don't need to use Generic V2 Webhook if the Slack extension will work.

"Generic V2 Webhook" pagerduty_extension_schema will create a new extension, but the extension won't work.

Also, I notice that we have older Slack extensions in Pagerduty with a "Slack (legacy webhook)" type 
"Slack (Legacy Webhook)" pagerduty_extension_schema errors out with the following:
"data.pagerduty_extension_schema.webhook: Unable to locate any extension schema with the name: Slack (Legacy Webhook)"

thanks,
Dusty

stmcallister:
This adds a new resource for Slack v2 connections in PagerDuty. Fixes #73,#94,#280,#357,#365

Test results:
```
TF_ACC=1 go test -run "TestAccPagerDutySlackConnection" ./pagerduty -v -timeout 120m
=== RUN   TestAccPagerDutySlackConnection_import
--- PASS: TestAccPagerDutySlackConnection_import (13.32s)
=== RUN   TestAccPagerDutySlackConnectionTeam_import
--- PASS: TestAccPagerDutySlackConnectionTeam_import (6.10s)
=== RUN   TestAccPagerDutySlackConnection_Basic
--- PASS: TestAccPagerDutySlackConnection_Basic (14.02s)
=== RUN   TestAccPagerDutySlackConnection_Team
--- PASS: TestAccPagerDutySlackConnection_Team (6.11s)
=== RUN   TestAccPagerDutySlackConnection_Envar
--- PASS: TestAccPagerDutySlackConnection_Envar (4.48s)
PASS
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	44.947s
```

### DIFF
--- a/pagerduty/resource_pagerduty_slack_connection.go
+++ b/pagerduty/resource_pagerduty_slack_connection.go
@@ -1,0 +1,309 @@
+package pagerduty
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func resourcePagerDutySlackConnection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourcePagerDutySlackConnectionCreate,
+		Read:   resourcePagerDutySlackConnectionRead,
+		Update: resourcePagerDutySlackConnectionUpdate,
+		Delete: resourcePagerDutySlackConnectionDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourcePagerDutySlackConnectionImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"source_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"source_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"source_type": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validateValueFunc([]string{
+					"service_reference",
+					"team_reference",
+				}),
+			},
+			"channel_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"channel_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"workspace_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SLACK_CONNECTION_WORKSPACE_ID", nil),
+			},
+			"notification_type": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validateValueFunc([]string{
+					"responder",
+					"stakeholder",
+				}),
+			},
+			"config": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"events": {
+							Type:     schema.TypeList,
+							Required: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"priorities": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+						"urgency": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validateValueFunc([]string{
+								"high",
+								"low",
+							}),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func buildSlackConnectionStruct(d *schema.ResourceData) (*pagerduty.SlackConnection, error) {
+	slackConn := pagerduty.SlackConnection{
+		SourceID:         d.Get("source_id").(string),
+		SourceName:       d.Get("source_name").(string),
+		SourceType:       d.Get("source_type").(string),
+		ChannelID:        d.Get("channel_id").(string),
+		ChannelName:      d.Get("channel_name").(string),
+		WorkspaceID:      d.Get("workspace_id").(string),
+		NotificationType: d.Get("notification_type").(string),
+		Config:           expandConnectionConfig(d.Get("config").(interface{})),
+	}
+	return &slackConn, nil
+}
+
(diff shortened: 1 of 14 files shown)


pagerduty/resource_pagerduty_slack_connection.go
```diff
@@ -0,0 +1,309 @@
...
+
+func expandConnectionConfig(v interface{}) pagerduty.ConnectionConfig {
+	c := v.([]interface{})[0].(map[string]interface{})
+
+	var config pagerduty.ConnectionConfig
+
+	config = pagerduty.ConnectionConfig{
+		Events:     expandConfigList(c["events"].([]interface{})),
+		Priorities: expandConfigList(c["priorities"].([]interface{})),
+		Urgency:    nil,
```
heimweh, line 234: Does this need to be nil? 🙂

pagerduty/resource_pagerduty_slack_connection.go
```diff
@@ -0,0 +1,309 @@
...
+		return retryErr
+	}
+
+	return nil
+}
+
+func resourcePagerDutySlackConnectionUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := &pagerduty.Config{
+		Token:   os.Getenv("PAGERDUTY_USER_TOKEN"),
+		BaseURL: "https://app.pagerduty.com",
```
heimweh, line 184: Do we want to make this a const? 🙂

pagerduty/resource_pagerduty_slack_connection.go
```diff
@@ -0,0 +1,309 @@
...
+	var items []interface{}
+
+	for _, i := range list {
+		items = append(items, i)
+	}
+
+	return items
+}
+
+func setClientConfig(c *pagerduty.Client) {
```
heimweh, line 276: Do we want to leverage this function to return an initialized client? 🙂

heimweh (commented):
Nice work! Can't wait to use this 🙂 :tada:

pagerduty/resource_pagerduty_slack_connection.go
jbfavre: Will it work with the EU service region ?
I can see the API URL is different depending on using US or EU region
Might be the case also for app.pagerduty.com ?

pagerduty/resource_pagerduty_slack_connection.go
stmcallister: Yeah, the API won't allow it to be a blank string. It's either a valid value or null.

pagerduty/resource_pagerduty_slack_connection.go
stmcallister: Good point @jbfavre! I confirmed with PagerDuty engineering that is the case. 🤔 Will need to figure out a better solution for setting baseURL.

Glen-Moonpig:
I now see the documentation says
`This resource requires a PagerDuty user-level API key set as the PAGERDUTY_USER_TOKEN environment variable.`
That is painful 😓

m1n9o:
@Glen-Moonpig I encountered the same problem. I was using PagerDuty Token with full access and was trying to create Slack V2 NG with terraform. It didn't work. Shouldn't we use Slack Token for it as well??? Any ideas??

cc: @stmcallister @jbfavre @heimweh

m1n9o:
> > @Glen-Moonpig I dont think so, I tried. It is only valid when PAGERDUTY_USER_TOKEN is set. Look [here](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/381/files#diff-9ec0b9b4671b295a357d9e17e4315116e1af16fa749d130f48c1df2af2cf7415R179).
> 
> Oh wow that is not good. We use module inputs for everything, no environment variables.

Yeah it's not good, but I think you can just export it in where terraform executed.
Actually, I wonder if they will improve this in near future or not.

rabidscorpio:
@jbfavre Can you explain why it has to be done at the user level? I've seen situations like this at other organizations where they backed off of requiring user level authentication specifically because people want to be able to manage configuration in an automated way without having to create a ghost user. Requiring a pagerduty user level api key means I have to waste a user just to be able to automate slack configuration. If I use a user-level key for an actual user and that user leaves the organization or has a security breach then I have to reset the pagerduty-slack connection and potentially (?) all of the services that use that connection.

This seems extremely short-sighted, especially in light of the fact that IaC is becoming so much more common now. Can Pagerduty please rethink this use case.

jbfavre:
@rabidscorpio not sure I'm the best candidate to do so, but I'll try anyway :wink: 
When trying to manually setup a Slack connection, you need to have both your Slack and PagerDuty accounts linked.
Which means PagerDuty got a OAuth token from Slack, on **your** behalf.
AFAIK, this is why you need a user token on terraform side: PagerDuty needs to be able to determine which OAuth token to use.

Now, why not using this OAuth token as an account token ?
I don't know exactly, but since it's a user token on Slack side, I think PagerDuty has to follow this workflow as well.
If not, there's a chances things will break whenever a user leaves the organisation.

But, again, I'm **not** a PagerDuty developer, so I might be wrong in my analysis.

Best,
JB

PR #382, opened 2021-09-06
integration: Validate that email integration has email

zecke:
Add validation that iff email integration is used that an email
address is specified. Make a plan only change to trigger an error
and a diff.

### DIFF
--- a/pagerduty/resource_pagerduty_service_integration.go
+++ b/pagerduty/resource_pagerduty_service_integration.go
@@ -17,6 +17,13 @@ func resourcePagerDutyServiceIntegration() *schema.Resource {
 		Read:   resourcePagerDutyServiceIntegrationRead,
 		Update: resourcePagerDutyServiceIntegrationUpdate,
 		Delete: resourcePagerDutyServiceIntegrationDelete,
+		CustomizeDiff: func(diff *schema.ResourceDiff, i interface{}) error {
+			t := diff.Get("type").(string)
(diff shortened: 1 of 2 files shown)


zecke:
@stmcallister @drastawi PTAL

pagerduty/resource_pagerduty_service_integration.go
```diff
@@ -17,6 +17,13 @@ func resourcePagerDutyServiceIntegration() *schema.Resource {
 		Read:   resourcePagerDutyServiceIntegrationRead,
 		Update: resourcePagerDutyServiceIntegrationUpdate,
 		Delete: resourcePagerDutyServiceIntegrationDelete,
+		CustomizeDiff: func(diff *schema.ResourceDiff, i interface{}) error {
+			t := diff.Get("type").(string)
+			if t == "generic_email_inbound_integration" && diff.Get("integration_email").(string) == "" {
+				return fmt.Errorf("integration_email attribute must be set")
```
drastawi, line 23:
```suggestion
				return fmt.Errorf("integration_email attribute must be set for an integration type generic_email_inbound_integration")
```

drastawi (approved):
Thanks @zecke

pagerduty/resource_pagerduty_service_integration.go
zecke: Updated the error message as well.

stmcallister (approved):
Awesome! Thank you!

PR #383, opened 2021-09-09
Add regex validation to start_time_of_day

drastawi:
Add validation for the start time format. It will prevent perma-diff issues with times like "9:00:00" rather than the PD-desired "09:00:00"


drastawi:
@stmcallister

stmcallister (approved):
Nice! 🎉  Thank you for adding this!

PR #384, opened 2021-09-09
Add support for provider `baseurl` new configuration option

jbfavre:
In order to be able to use Pagerduty provider to manage EU platform deployment, we need to be able to define a custom API URL [0]

As gar as I can tell,  heimweh/go-pagerduty client use baseURL which is already configurable.
I tried to pass this configuration option when setting up heimweh/go-pagerduty.

@stmcallister please be indulgent, this is one of my first contribution in Go language :wink: 
I surely missed many things.
Also , I don't know how to test this feature. Advice will be very welcomed

[0]: https://support.pagerduty.com/docs/service-regions


stmcallister:
Well done! This project was my first using the go language as well! Code-wise, I think it's great. However, I think I want to go a slightly different direction. Rather than having folks set the URLs directly I think we should just provide a `service_region` field to the `provider` tag. Then the provider will figure out which URLs to use. At least that's my thoughts right now. 

From a user perspective, do you think that would be a cleaner approach?

jbfavre:
I understand and agree that having only one configuration option would help to avoid misbehaviours.

>  think we should just provide a service_region field to the provider tag

Do you mean, like:
```
provider "pagerduty" {
  token                = xXxXxX
  service_region = "EU"
}
```

The only caveat I can think of is that having "hard coded" URLs would require a provider release for each new service region opened. It's not that I expect many service region opening, but still I think we should look for the easiest solution.

Maybe dynamically building URLs would help. Kind of:
```
if service_region == "" or service_region == "us"
then
  ApiUrl = "https://api.pagerduty.com"
  AppUrl = "https://app.pagerduty.com"
else
  ApiUrl = "https://api." + service_region + ".pagerduty.com"
  AppUrl = "https://app." + service_region + ".pagerduty.com"
```

pagerduty/provider.go
```diff
@@ -99,7 +105,17 @@ func handleNotFoundError(err error, d *schema.ResourceData) error {
 }
 
 func providerConfigure(data *schema.ResourceData, terraformVersion string) (interface{}, error) {
+	var ServiceRegion = data.Get("service_region").(string)
```
stmcallister, line 108: Let's just always set the value of `ServiceRegion` to lowercase.

stmcallister (changes requested):
Great work! Thank you for this! Just a couple changes before approving.

pagerduty/provider.go
jbfavre: Done

jbfavre:
Indeed, I broke some tests in my branch :thinking: 
```
=== RUN   TestAccPagerDutySlackConnectionTeam_import
    import_pagerduty_slack_connection_test.go:42: Step 1/2 error: Error running apply: exit status 1
        
        Error: POST API call to https://api.pagerduty.com/integration-slack/workspaces/T02JBV9PANA/connections failed 404 Not Found. Code: 2100, Errors: <nil>, Message: Not Found
        
          with pagerduty_slack_connection.foo,
          on terraform_plugin_test.tf line 5, in resource "pagerduty_slack_connection" "foo":
           5: 		resource "pagerduty_slack_connection" "foo" {
        
--- FAIL: TestAccPagerDutySlackConnectionTeam_import (125.60s)
```

Ill have an look at it :eyes:

PR #385, opened 2021-09-10
Validate start time of week in a schedule

drastawi:
This can catch the errors when a user assumes 0 is the first day of the week.


stmcallister (approved):
Very useful! Thanks!! 👍

PR #386, opened 2021-09-12
Ensure start_day_of_week is not specified for a daily schedule restriction

drastawi:
This can prevent permanent diffs in cases start day of the week is defined on a daily restriction. This can help in the case that the plugin user has a misconception about what a "daily_restriction" means.


stmcallister (approved):
Looks good! Thank you for this! 👍

PR #387, opened 2021-09-13
Ensure general urgency rule is not set for an urgency rule of type support hours

drastawi:
If not checked at the plan level, it will lead to a parma diff


stmcallister (approved):
Great stuff!! Thank you for the contribution! 🎉

PR #395, opened 2021-09-17
Set constraints on Escalation Policy num_loops

drastawi:
All values outside of that range are rejected bu PagerDuty API


stmcallister (approved):
Looks great! Thank you! 👍 🎉

PR #402, opened 2021-10-06
Adding Tag and Tag Assignment resources. Also Tag Data Source.

stmcallister:
Adding `pagerduty_tag` and `pagerduty_tag_assignment` resources as well as `pagerduty_tag` data source. Here are the test results for those additions.

```
$ TF_ACC=1 go test -run "TestAccPagerDutyTagAssignment" ./pagerduty -v -timeout 120m 
=== RUN   TestAccPagerDutyTagAssignment_import
--- PASS: TestAccPagerDutyTagAssignment_import (8.74s)
=== RUN   TestAccPagerDutyTagAssignment_User
--- PASS: TestAccPagerDutyTagAssignment_User (8.58s)
=== RUN   TestAccPagerDutyTagAssignment_Team
--- PASS: TestAccPagerDutyTagAssignment_Team (6.43s)
=== RUN   TestAccPagerDutyTagAssignment_EP
--- PASS: TestAccPagerDutyTagAssignment_EP (9.90s)
PASS
ok      github.com/terraform-providers/terraform-provider-pagerduty/pagerduty   34.224s
```

```
$ TF_ACC=1 go test -run "TestAccDataSourcePagerDutyTag" ./pagerduty -v -timeout 120m
=== RUN   TestAccDataSourcePagerDutyTag_Basic
--- PASS: TestAccDataSourcePagerDutyTag_Basic (3.29s)
PASS
ok      github.com/terraform-providers/terraform-provider-pagerduty/pagerduty   3.861s
```


pagerduty/resource_pagerduty_tag_assignment.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if err != nil {
+		return []*schema.ResourceData{}, fmt.Errorf("error importing pagerduty_tag_assignment: %s", err.Error())
+	}
+	var foundTag *pagerduty.Tag
+	// loop tags and find matching ID
+	for _, tag := range tagResponse.Tags {
+		if tag.ID == tagID {
+			// create tag_assignment id using the entityID.tagID as PagerDuty API does not return one
+			assignmentID := fmt.Sprintf("%v.%v", entityID, tagID)
```
heimweh, line 174: Would it make sense to have a function that returns the ID in the expected format,
so we can easily re-use it in other places? 🙂 

```go
func assignmentID(entityID, tagID string) string {
  return fmt.Sprintf("%v.%v", entityID, tagID)
} 
```

pagerduty/resource_pagerduty_tag_assignment_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+// func init() {
```
heimweh, line 14: Do we want sweepers for the tag assignments? 🙂

pagerduty/resource_pagerduty_tag_assignment.go
```diff
@@ -0,0 +1,189 @@
...
+func resourcePagerDutyTagAssignmentDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*pagerduty.Client)
+
+	assignment := buildTagAssignmentStruct(d)
+	assignments := &pagerduty.TagAssignments{
+		Remove: []*pagerduty.TagAssignment{assignment},
+	}
+	log.Printf("[INFO] Deleting PagerDuty tag assignment with tagID %s for entityID %s", assignment.TagID, assignment.EntityID)
+
+	retryErr := resource.Retry(10*time.Second, func() *resource.RetryError {
```
heimweh, line 135: We could also return the result of `Retry()` here I think like we do in the `Read()` func 🙂 
```go
	return resource.Retry(10*time.Second, func() *resource.RetryError {
		if _, err := client.Tags.Assign(assignment.EntityType, assignment.EntityID, assignments); err != nil {
			if isErrCode(err, 400) || isErrCode(err, 429) {
			}

			return resource.NonRetryableError(err)
		} else {
			d.SetId("")
		}
		return nil
	})

```

website/docs/d/tag.html.markdown
```diff
@@ -0,0 +1,53 @@
+---
```
heimweh, line 1: I think this page just needs to be adjusted to match the `tag` data source 🙂

heimweh (commented):
Nice work @stmcallister :tada: I left a few inline comments 🙂

website/docs/d/tag.html.markdown
stmcallister: 🤦

PR #405, opened 2021-10-09
Add validation on schedule layer turn rotation length

drastawi:
For values outside of this range, the API will respond with a 400


drastawi:
@stmcallister any chance we could include this one in with the upcoming release?

PR #417, opened 2021-11-01
update vendor docs to update inaccurate example

satyapramodh:
# Why
the example for `pagerduty_service_integration` within `data.pagerduty_vendor` had both type and vendor defined. This resulted in an error for me when i copied over the code and modified it for my usage.

# Fix
As per [service_integration docs](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service_integration#argument-reference), type cannot be used when vendor is defined. So in accordance to that, i modified the doc.


stmcallister (approved):
Good catch! Thank you! 👍 🎉

PR #425, opened 2021-11-30
pagerduty_service_integration: support emails that are only known after apply

Yannic:
This change allows to use outputs from other resources (e.g. `random_id`) in
the `integration_email` field.

Example usage:

```terraform
resource "random_id" "foo" {
  keepers = {
    prefix = "foo"
  }

  byte_length = 8
}

resource "pagerduty_service_integration" "foo" {
  name = "email"

  service = pagerduty_service.foo.id

  type              = "generic_email_inbound_integration"
  integration_email = "${random_id.foo.keepers.prefix}-${random_id.foo.hex}@domain.pagerduty.com"
}
```

### DIFF
--- a/pagerduty/resource_pagerduty_service_integration.go
+++ b/pagerduty/resource_pagerduty_service_integration.go
@@ -2,6 +2,7 @@ package pagerduty
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -12,6 +13,10 @@ import (
 	"github.com/heimweh/go-pagerduty/pagerduty"
 )
 
+const (
+	errEmailIntegrationMustHaveEmail = "integration_email attribute must be set for an integration type generic_email_inbound_integration"
+)
+
 func resourcePagerDutyServiceIntegration() *schema.Resource {
 	return &schema.Resource{
 		Create: resourcePagerDutyServiceIntegrationCreate,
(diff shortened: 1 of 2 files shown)


pagerduty/resource_pagerduty_service_integration.go
```diff
@@ -111,13 +116,20 @@ func buildServiceIntegrationStruct(d *schema.ResourceData) *pagerduty.Integratio
 		}
 	}
 
-	return serviceIntegration
+	if serviceIntegration.IntegrationEmail == "" {
```
drastawi, line 119: @Yannic this is not necessarily an email integration? Just like in customize diff you should also check for `t == "generic_email_inbound_integration"`. I'm actually surprised the tests did not fail, perhaps we should modify the existing tests to confirm this will not fail if service integration is of another type?

drastawi (changes requested):
Hey @Yannic, this is very useful and we should merge, but I think you just need one small fix

pagerduty/resource_pagerduty_service_integration.go
Yannic: done

drastawi (approved):
lgtm

Yannic:
@stmcallister any chance you could take a look at this? this issue is blocking us from fully adopting the PagerDuty terraform provider

stmcallister (approved):
@Yannic this looks awesome! Thank you for doing this! Is this not something we want to document? I could go either way. Feels like the "email known after apply" scenario is pretty specific?

Yannic:
@stmcallister thanks for your review! I'm not sure I follow what should be documented. It's pretty common in terraform for values to be only known after apply (not only for email with `random_id`), so I'd say once this is merged there are less pitfalls and less things to document. I was pretty surprised when seeing the current behavior.

If there are other fields you know of that can't be known after apply today, we should probably fix them as well.

Padorax:
Hey @Yannicm this looks great! However, just wondering have you tested when the integration email address is explicitly set to an empty string (or even not set at all), is this customizeDiff validation still able to detect and error out? Thanks!

PR #435, opened 2021-12-22
Validate user and schedule reference in escalation policy targets

drastawi:
This prevents a use of "user" and "schedule" types in escalation policy target as well as other invalid values.

Notably a "user" type is being returned as "user_reference" by the PagerDuty API, so it is causing a perma-diff in terraform plan like below:

```
      ~ rule {
            id                          = "PD..."
            # (1 unchanged attribute hidden)

          ~ target {
                id   = "PH..."
              ~ type = "user_reference" -> "user"
            }
        }

```

### DIFF
--- a/pagerduty/resource_pagerduty_escalation_policy.go
+++ b/pagerduty/resource_pagerduty_escalation_policy.go
@@ -66,6 +66,10 @@ func resourcePagerDutyEscalationPolicy() *schema.Resource {
 										Type:     schema.TypeString,
 										Optional: true,
 										Default:  "user_reference",
+										ValidateFunc: validateValueFunc([]string{
+											"user_reference",
+											"schedule_reference",
+										}),
 									},
 									"id": {
 										Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


drastawi:
@stmcallister this will also require some docs update, but I want to make sure we are ok with the change as likely many people use the "users" and "schedules" in their targets.

stmcallister:
@drastawi Yeah, this change makes sense. And, I agree with your suggestion. Will you also update the documentation on the escalation policy resource as well?

drastawi:
@stmcallister good catch, added

stmcallister:
Thanks @drastawi ! Docs look great! Although, now I'm getting merge conflicts between this branch and the current version of master. Will you get the latest version of master and fix those conflicts? Thanks!

drastawi:
Ok, all rebased. Let me know if you still see any problems.

stmcallister (approved):
Awesome! Thank you! 👍 🎉

PR #436, opened 2021-12-22
Example code update for newer versions of terraform

### LINKED ISSUE #434: Vanilla example usage needs to include the terraform block
Hi there,

Looks like the vanilla example usage no longer works with the newest version of terraform.

### Terraform Version
1.1.2

### Affected Resource(s)
provider

```hcl
provider "pagerduty" {
  token = "u+..."
}
```

### Debug Output
Initializing provider plugins...

### Panic Output
```
- Finding latest version of hashicorp/pagerduty...
╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/pagerduty: provider registry registry.terraform.io does not have a provider named registry.terraform.io/hashicorp/pagerduty
│ 
│ Did you intend to use pagerduty/pagerduty? If so, you must specify that source address in each module which requires that provider. To see which modules are currently depending on hashicorp/pagerduty, run
│ the following command:
│     terraform providers
╵
```

### Expected Behavior
Should initialize

### Actual Behavior
Errors out

### Steps to Reproduce
1. `terraform init`

### Important Factoids

This should be added to the example:
```
terraform {
  required_providers {
    pagerduty = {
      source  = "pagerduty/pagerduty"
    }
  }
}
```

drastawi:
fixes #434 

Updating the sample project to include the `terraform {} ` block which had been required for a while now when using non-HashiCorp providers. Also added `.gitignore` for the terraform lock files that are now generated with any terraform project on initialization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 example.tf
 terraform.tfplan
 terraform.tfstate
+.terraform.lock.hcl
 bin/
 dist/
 modules-dev/
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -16,6 +16,14 @@ Use the navigation to the left to read about the available resources.
 
 ```hcl
 # Configure the PagerDuty provider
+terraform {
+  required_providers {
+    pagerduty = {
+      source  = "pagerduty/pagerduty"
+    }
+  }
+}
+
 provider "pagerduty" {
   token = var.pagerduty_token
 }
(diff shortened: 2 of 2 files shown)


stmcallister (changes requested):
Oh, this is a great addition! Will you also add a `version` field to the `pagerduty` block?

stmcallister (approved):
👍 Awesome! Thank you! 🎉

PR #433, opened 2021-12-22
Validate "duration_seconds" for a layer restriction

drastawi:
Just a small validation addition for a common confusion of daily restriction being understood as "restrict a for a full day".

PR #439, opened 2021-12-24
schedule: Validate daily/weekly restrictions

zecke:
Do not allow a daily restriction that spans a full 24*time.Hour
and similar for weekly restrictions. The PD API rejects 24*time.Hour
and 24*7*time.Hour for the API.


stmcallister (changes requested):
Oooo! I like this solution! It's really readable AND you have tests!!! 😍 I just merged #433 which accomplishes the same thing, but I would also like this one because of readability and the included tests. However, when I run the tests with the newly merged code the test fails with the following message. If you could fix this I would love to merge it in to the project.

```
=== RUN   TestAccPagerDutySchedule_BasicWeek
    resource_pagerduty_schedule_test.go:216: Step 3/3, expected an error with pattern, no match on: Error running pre-apply refresh: exit status 1
        
        Error: expected layer.0.restriction.0.duration_seconds to be in the range (1 - 604799), got 604800
        
          with pagerduty_schedule.foo,
          on terraform_plugin_test.tf line 23, in resource "pagerduty_schedule" "foo":
          23:       duration_seconds  = 604800
        
--- FAIL: TestAccPagerDutySchedule_BasicWeek (12.00s)
```

PR #454, opened 2022-02-03
Increase Retry Time on Data Sources

stmcallister:
Some users are still running in to `429: Too Many Requests` errors from the PagerDuty API, specifically on Data Sources. There are also a few resources with increased retry times. And a few API calls in the Escalation Policy Resource and the Service Dependency Resource that Retry was added.

### DIFF
--- a/pagerduty/resource_pagerduty_service_dependency.go
+++ b/pagerduty/resource_pagerduty_service_dependency.go
@@ -127,7 +127,7 @@ func resourcePagerDutyServiceDependencyAssociate(d *schema.ResourceData, meta in
 	log.Printf("[INFO] Associating PagerDuty dependency %s", serviceDependency.ID)
 
 	var dependencies *pagerduty.ListServiceDependencies
-	retryErr := resource.Retry(30*time.Second, func() *resource.RetryError {
+	retryErr := resource.Retry(5*time.Minute, func() *resource.RetryError {
 		if dependencies, _, err = client.ServiceDependencies.AssociateServiceDependencies(&input); err != nil {
 			if isErrCode(err, 404) {
 				return resource.RetryableError(err)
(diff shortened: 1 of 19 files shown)


pagerduty/resource_pagerduty_service_dependency.go
```diff
@@ -160,26 +160,30 @@ func resourcePagerDutyServiceDependencyDisassociate(d *schema.ResourceData, meta
...
 
-	// loop serviceRelationships until relationship.IDs match
-	for _, rel := range depResp.Relationships {
-		if rel.ID == d.Id() {
-			foundDep = rel
-			break
+	// listServiceRelationships by calling get dependencies using the serviceDependency.DependentService.ID
+	retryErr := resource.Retry(5*time.Minute, func() *resource.RetryError {
+		if dependencies, _, err := client.ServiceDependencies.GetServiceDependenciesForType(dependency.DependentService.ID, dependency.DependentService.Type); err != nil {
+			if isErrCode(err, 404) || isErrCode(err, 500) || isErrCode(err, 429) {
```
drastawi, line 168: @stmcallister, would be a good idea to centralize the logic of what is a returnable error? I'm not sure if it is worth it to retry on 400 at all for example. I can see the benefit of increasing the retry to 5min for 429 and maybe 503/504 errors, but it might substantially hurt the experience if there is a real 404 or 400.

stmcallister: @drastawi Thanks for the feedback! The tricky part is that in some race condition situations we do want to retry on a 404 🤔 Good point on the centralizing of the logic.

pagerduty/resource_pagerduty_service_dependency.go
```diff
@@ -193,17 +197,17 @@ func resourcePagerDutyServiceDependencyDisassociate(d *schema.ResourceData, meta
 	input := pagerduty.ListServiceDependencies{
 		Relationships: r,
 	}
-	retryErr := resource.Retry(30*time.Second, func() *resource.RetryError {
+	retryErr = resource.Retry(5*time.Minute, func() *resource.RetryError {
 		if _, _, err = client.ServiceDependencies.DisassociateServiceDependencies(&input); err != nil {
-			if isErrCode(err, 404) {
+			if isErrCode(err, 404) || isErrCode(err, 429) {
 				return resource.RetryableError(err)
```
drastawi, line 203: @stmcallister I am a bit concerned the end users might have to wait for 5 min to find out a resource already exists or had already been deleted manually which are not that uncommon. 

One solution that seems fairly easy to implement at a glance would be to add a max timeout to the retryable error as an extra optional parameter

```suggestion
                        if isErrCode(err, 429) {
				return resource.RetryableError(err)
			} else if  isErrCode(err, 404) {
				return resource.RetryableErrorWithMaxTimeout(err, 15 * time.Second)
			}
```

dobs (commented):
Overall LGTM once merge conflicts are resolved -- can give the :+1: once fixed.

To @drastawi's feedback there's some room for potential consolidation and adjustment but I'd be fine with tackling that in a separate pass, possibly including moving values into the `Config`.

dnck:
Over here, we're still seeing 429s with PD provider 3.2.0...

Issue #453, opened 2022-02-02
panic in dataSourcePagerDutyVendorRead


ryanking:
### Terraform Version

1.1.3

### Affected Resource(s)

- data_source_pagerduty_vendor

### Terraform Configuration Files
```hcl
data "pagerduty_vendor" "datadog" {
  name = "Datadog"
}
```

### Debug Output

I don't have this because the error is transient and I didn't have debug logging on when it happend.

### Panic Output

https://gist.github.com/436274ff78e5c5a893f82948427424c0

### Steps to Reproduce

I have not been able to reproduce this yet.

### Important Factoids

Nothing out of the ordinary with this code. We recently upgraded to the latest provider version.

dobs:
Heads-up that this may be addressed as part of #458 and released as part of 2.3.0.

Issue #447, opened 2022-01-17
The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call.


katkasian:
### Terraform Version
Terraform v1.1.3
on darwin_amd64
+ provider registry.terraform.io/pagerduty/pagerduty v2.2.1

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_ruleset_rule

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
https://drive.google.com/file/d/1xuDkQClruBJlSMxwf-dFCj86jHj1fGXW/view?usp=sharing

### Debug Output
https://gist.github.com/katkasian/c031f41ac5fa0ea6ad9729f591171863

### Panic Output
NA

### Expected Behavior
When creating a module to create rulesets, I expect the apply to be successful.

### Actual Behavior
Encountering "The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details." error, provider crashes. On PagerDuty side, ruleset gets created, but not individual rules.

### Steps to Reproduce
1. Download the configuration files
2. Initiate a terraform repo in the `Archive` folder.
3. Create / copy PagerDuty API key
4. Run `terraform apply`, provide the key when prompted.

### Important Factoids
Rulesets are managed using a module (contained in ruleset folder of the zip file)

NargiT:
I have also some issues with this version.
How many resources do you create ? 

```
│ Error: Request cancelled
│ 
│   with pagerduty_service.xxxxx,
│   on auto-service-xxxxx.tf line 4, in resource "pagerduty_service" "xxxxx":
│    4: resource "pagerduty_service" "xxxxx" {
│ 
│ The plugin.(*GRPCProvider).UpgradeResourceState request was cancelled.
╵

Stack trace from the terraform-provider-pagerduty_v2.1.1 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0xf27f34]

goroutine 463 [running]:
github.com/terraform-providers/terraform-provider-pagerduty/pagerduty.resourcePagerDutyServiceRead.func1(0x2)
	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty/resource_pagerduty_service.go:379 +0x94
[...]

Issue #440, opened 2022-01-04
Nil pointer dereference panic during cancellation in v2.2.1


jcogilvie:
Disclaimer:  this is an issue caused by another upstream issue.  An upstream provider got a 503 from the service it talks to, which caused things to be canceled.  Apparently 2.2.1 does not handle this cancellation gracefully.

Since I cannot reproduce the 503 for you, the bug report is going to be a bit limited.

### Terraform Version
1.0.11

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_user

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "pagerduty_user" "user" {
  count = var.create_pagerduty_user ? 1 : 0

  name  = "${var.first_name} ${var.last_name}"
  email = var.employee_email
  role  = var.pagerduty_role

  lifecycle {
    ignore_changes = [
      job_title,
    ]
  }
}
```

### Debug Output
No debug input since I can't readily repro with it on due to the transient nature, but here's an excerpt.  Each of these occurs many times (since we have many users), except for the one emitted by the provider that got a 503.

```
╷
│ Error: Request cancelled
│ 
│   with data.pagerduty_schedule.sr_leadership_schedule,
│   on global-pagerduty.tf line 1, in data "pagerduty_schedule" "sr_leadership_schedule":
│    1: data "pagerduty_schedule" "sr_leadership_schedule" {
│ 
│ The plugin.(*GRPCProvider).ValidateDataResourceConfig request was
│ cancelled.
╵
╷
│ Error: client failed: client error 503: 
│ 
│   with provider["registry.local/broken"],
│   on providers.tf line 148, in provider "broken":
│  148: provider "broken" {
│ 
╵
╷
│ Error: Request cancelled
│ 
│   with module.architecture_group.data.pagerduty_vendor.datadog,
│   on modules/group/pagerduty.tf line 39, in data "pagerduty_vendor" "datadog":
│   39: data "pagerduty_vendor" "datadog" {
│ 
│ The plugin.(*GRPCProvider).ValidateDataResourceConfig request was
│ cancelled.
╵
╷
│ Error: Request cancelled
│ 
│   with module.core_services_group.data.pagerduty_vendor.datadog,
│   on modules/group/pagerduty.tf line 39, in data "pagerduty_vendor" "datadog":
│   39: data "pagerduty_vendor" "datadog" {
[...]

Issue #421, opened 2021-11-18
Segmentation violation error instead of a "wrong credentials" message when trying to apply with an invalid token


christianbrideau:
### Terraform Version
Terraform v1.0.11
on linux_amd64

### Affected Resource(s)
- pagerduty_user
- pagerduty_team
- Probably others

The problem occurs when trying to apply using a bad token.

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    pagerduty = {
      source = "PagerDuty/pagerduty"
      version = "2.1.1"
    }
  }
}

provider "pagerduty" {
  token = "any_incorrect_token"
}

resource "pagerduty_user" "muser" {
      name  = "Mister User"
      email = "muser@company.com"

}
```

### Debug Output
https://gist.github.com/christianbrideau/7af93697c93092e7b6e36a55bcb12c62

### Panic Output
No crash.log file created.

### Expected Behavior
Expecting a message of type "access denied" or "wrong credentials" when a bad token is used (missed a character when copying/pasting).

### Actual Behavior
Getting a segmentation fault error.  I can see the access denied message, but only when turning on debugging.  This sent me on a wild goose chase to find out why it segfaulted.  Tried different versions of Terraform, the provider and operating system, when it was just me not being able to do a copy/paste correctly.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:

1. Use the config file from above.
2. Use a wrong token.
3. `terraform apply`

### Important Factoids
This was done on a sandbox account.  

### References
No

dobs:
This may now be addressed as part of #458, released as part of 2.3.0.

In this particular case the error message should now be surfaced rather than causing a segfault.

PR #458, opened 2022-02-08
Safer HTTP client initialization and usage.

dobs:
This PR has two major components (corresponding to the two commits):

- Updating API client creation to only initialize once rather than every time a client is referenced.
- Surfacing client errors whenever an API client is retrieved from the config.

By exposing new errors this change may break some expectations around what errors might be returned by the provider, but hopefully in a way that's useful (i.e. before it would segfault, now it'll hopefully return something more informative).

The one gotcha would be if client re-initialization was intentional, but hopefully that isn't the case. And one optimization might be to explicitly initialize the client at config time, after which we don't need error handling around `Client()` calls.

This PR likely addresses the segfaults mentioned in issues:

- #394 
- #421 
- #440 
- #447
- #453

### DIFF
--- a/pagerduty/config.go
+++ b/pagerduty/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"sync"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
 	"github.com/heimweh/go-pagerduty/pagerduty"
@@ -11,6 +12,8 @@ import (
 
 // Config defines the configuration options for the PagerDuty client
 type Config struct {
+	mu sync.Mutex
+
 	// The PagerDuty API URL
 	ApiUrl string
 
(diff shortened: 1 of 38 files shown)


pagerduty/config.go
```diff
@@ -40,8 +46,16 @@ Please see https://www.terraform.io/docs/providers/pagerduty/index.html
 for more information on providing credentials for this provider.
 `
 
-// Client returns a new PagerDuty client
+// Client returns a PagerDuty client, initializing when necessary.
 func (c *Config) Client() (*pagerduty.Client, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
gempesaw, line 52: haven't used mutexes before in Go, but pleasantly surprised how simple this looks! that's all it takes to handle "a bunch of different objects ask for a client at the same time, and we only want to make one, and have all my callers wait until i make that and give the single one to everyone" ? :o cooool

stmcallister (approved):
Thanks @dobs !! I think this will be a great improvement to user experience.

Issue #396, opened 2021-09-20
Unable to remove pagerduty maintenance window using terraform 


msa21:
Terraform version we are using 

 terraform -v
Terraform v0.12.24

Error: DELETE API call to https://api.pagerduty.com/maintenance_windows/P7IXGF5 failed 405 Method Not Allowed. Code: 5003, Errors: <nil>, Message: Cannot cancel a maintenance window from the past

We are trying to remove the maintenance window while it is not expired  but it is still giving the same error .

sbchisholm:
We've also run into this as well.
```
terraform -v
Terraform v1.1.6
on linux_amd64
...
+ provider registry.terraform.io/pagerduty/pagerduty v2.3.0
```

After the maintenance window has passed it can no longer be updated or destroyed:

Update error:
```
│ Error: PUT API call to https://api.pagerduty.com/maintenance_windows/ABCDEFG failed 400 Bad Request. Code: 2001, Errors: [Start time cannot be modified for maintenance windows that have already started End time already passed and can not be modified], Message: Invalid Input Provided
```

Destroy error:
```
│ Error: DELETE API call to https://api.pagerduty.com/maintenance_windows/ABCDEFG failed 405 Method Not Allowed. Code: 5003, Errors: <nil>, Message: Cannot cancel a maintenance window from the past
```

kartsm:
I may be running into this as well.  I am using Terraform 1.0.11 with version 2.3.0 of the PagerDuty provider.

I successfully created maintenance windows for a couple of services.  I am providing what source code that I can, trying to highlight only the most relevant portions.  Timestamps turn out to be important to this report.  Names have been changed to protect the innocent. 😇 

In `main.tf`:
```hcl
module "pagerduty_critical" {
  source = "/path/to/xxx-pagerduty"
  /* some settings omitted... */
  maint_window_start       = var.pagerduty_maint_window_start
  maint_window_end         = var.pagerduty_maint_window_end
}
```
In `variables.tf`:
```hcl
variable "pagerduty_maint_window_end" {
  /* ... description omitted ... */
  type        = string
  default     = "2022-04-10T23:59:59-04:00"
  /* ... validation omitted ... */
}

variable "pagerduty_maint_window_start" {
  /* ... description omitted ... */
  type        = string
  default     = "2022-03-01T00:00:00-05:00"
  /* ... validation omitted ... */
}
```
In `xxx-pagerduty/variables.tf`:
```hcl
[...]

PR #467, opened 2022-03-02
Delete and re-create maintenance window if start or end times change

sbchisholm:
Proposed solution for the 400 error seen in issue #396, when attempting to update the start and end times of a maintenance window.


stmcallister:
I like this idea! Would you be able to write a test for this scenario and add it to the acceptance tests?

sbchisholm:
> I like this idea! Would you be able to write a test for this scenario and add it to the acceptance tests?

Sorry for taking forever to reply. It wasn't immediately clear where I could add test coverage for this in the existing test suite, I'll have another chance to get back to this soon.

In the meantime we've just been working around the issue by using `terraform state rm ...` to remove the past window from the state.

Issue #457, opened 2022-02-08
Feature Request: Add a data source for a existing response play


jjm:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

1.1.5

### Affected Resource(s)

n/a

### Terraform Configuration Files

The expected way of referencing this would be:

```hcl
data "pagerduty_response_play" "example" {
  name = "My Response Play,"
}
```

### Debug Output
n/a, data source doesn't exist.

### Panic Output
n/a

### Expected Behavior
It should be possible to get PagerDutyID for an existing response play, as some features (e.g. create a Zoom meeting?), can't be configured in Terraform.

### Actual Behavior
I can't get the reference for an existing response play and then assign it to a service via Terraform.

### Steps to Reproduce
n/a

### Important Factoids

- It's possible to list all response-plays via: https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE2NQ-list-response-plays

### References

- Adding support to response play's for a service has been requested in #400

jjm:
I'm happy to work on this, as we want to automate setting up response plays on services via Terraform.

PR #470, opened 2022-03-12
Add pagerduty_response_play datasource

jjm:
- Add a pagerduty_response_play data source (fixes: #457)

This would be used as follows:

```terraform
data "pagerduty_response_play" "this" {
  name = "Response Play"
  from = data.pagerduty_user.this.email
}
```

Then it'll be possible to reference the response play via `data.pagerduty_response_play.this.id`, which would be needed for response adding a webUI created response play to a service, as some features are webUI only (e.g. Zoom conference bridge).

**Questions:**
- Should the from be on the data source? Or a default added to the provider?

**ToDo**
- [x] Add docs.
- [ ] Add tests.

### DIFF
--- a/pagerduty/data_source_pagerduty_response_play.go
+++ b/pagerduty/data_source_pagerduty_response_play.go
@@ -73,7 +73,6 @@ func dataSourcePagerDutyResponsePlayRead(d *schema.ResourceData, meta interface{
 
 		d.SetId(found.ID)
 		d.Set("name", found.Name)
-		d.Set("from", found.FromEmail)
 
 		return nil
 	})
--- a/pagerduty/data_source_pagerduty_response_play.go
+++ b/pagerduty/data_source_pagerduty_response_play.go
@@ -1,0 +1,80 @@
+package pagerduty
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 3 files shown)


jjm:
@stmcallister I know I've not progressed this for some time, but do you have any idea if the `from` for this API call should be on the provider or on the data source? That's part of the reason that I've not added docs or the completed tests.

stmcallister:
@jjm The way we handled fields like this is to have the field in the resource (or data source, in this case) and then provide a `DefaultFunc` value where an environment variable can be used if set. A good example is the [resource_pagerduty_slack_connection](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/898f8d1b73d446762ab234bebca18250585c2bbb/pagerduty/resource_pagerduty_slack_connection.go#L53) and with the workspace_id.

I would recommend going this route for consistency. Thanks!

jjm:
@stmcallister Apologies for the really late reply, I was just looking at doing this and it seems that [response_play](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/response_play) already has a `from` field defined on the resource as I've done.

I'm wondering if it should be constant with `response_play` resource?

jjm:
I've added the missing docs, just need to look into adding the missing tests.

jjm:
I don't have much time for open source work currently and we have stopped using response plays. So I'm just going to close this off.

PR #472, opened 2022-03-16
[DO NOT MERGE] Event Orchestration Resource proposal

alenapan:


stmcallister:
I believe this was covered with #512. Closing.

PR #473, opened 2022-03-21
Validate time zones

drastawi:
We can leverage go's time package to validate time zones are valid.

On an invalid time zone the provider will fail with the message:
```
Failed to load location: unknown time zone America/nyc
```


stmcallister (approved):
Cool! Thanks @drastawi!! 👍

PR #475, opened 2022-03-24
Phone numbers starting with a 0 are not supported

drastawi:
There are still a lot of phone numbers not supported by PagerDuty, a particularly large set of them is numbers starting with 0, particularly common in UK and Australia. There is currently no timeline when this might be fixed. Until the numbers are enabled, we should validate them at the provider level.

In addition to leading zero, we know only the digits-only format is supported


stmcallister (approved):
Good call! Thanks for this! 🙏 👍

PR #479, opened 2022-04-07
Fix permadiff in email_parser with type regex & minor docs update

bohdanborovskyi-ma:
This PR address the permadiff issue which occurs in `email_parser` fields on the `pagerduty_service_integration` resource when `type = "regex"` is selected. 
Also includes minor documentation improvements for same area.

P.s. I wasn't able to verify the changes yet as I am pretty-new to Golang and just compared codebase in flattenEmailParsers() & expandEmailParsers() functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 2.4.1 (April 22, 2022)
+IMPROVEMENTS:
+* `resource/pagerduty_user_notification`: Create user/notification rule: allow using existing ones ([#482](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/482))
+* `resource/pagerduty_schedule`: Enforce 0 second time restriction on schedules ([#483](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/483))
+
+BUG FIXES:
+* Embed time zone data into the provider's binary([#478](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/478))
+* Documentation fix: update all broken links pointing to PagerDuty API documentation ([#464](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/464))
+
+
 ## 2.4.0 (April 1, 2022)
 FEATURES:
 * `resource/pagerduty_service_integration`: Add Email Filters to Service Integrations ([#468](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/468))
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -6,7 +6,7 @@ PKG_NAME=pagerduty
 default: build
 
 build: fmtcheck
-	go install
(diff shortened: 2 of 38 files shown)


bohdanborovskyi-ma:
@stmcallister @drastawi Could you please review & release this PR?

Right now, we're using a workaround in our internal modules (ignoring the changes in `regex` field), but seems like that workaround causing another issue.

When TF detects changes in any of other field within `pagerduty_service_integration` resource, resulting request during `terraform apply` fails as payload sent to PD API endpoint became invalid. We receive 400 Error due to regex field being silently omitted by TF (due to `ignore_changes` lifecycle parameter).

So we ended up with 2 mutually exclusive issues:
- didn't use `ignore_changes` and receive constant drift due to bug in flattenEmailParsers() function
- use `ignore_changes` and receive 400 API errors when trying to update other attributes of integration resource

We're heavily relying on those email parsing functionality and we'd like to see that fix being merged into upstream provider as soon as possible. Thanks in advance

pagerduty/resource_pagerduty_service_integration.go
```diff
@@ -519,12 +519,21 @@ func flattenEmailParsers(v []*pagerduty.EmailParser) []map[string]interface{} {
...
+					"regex":      ve.Regex,
+				}
+			} else {
+				extractor := map[string]interface{}{
+					"type":         ve.Type,
+					"value_name":   ve.ValueName,
+					"part":         ve.Part,
+					"starts_after": ve.StartsAfter,
+					"ends_before":  ve.EndsBefore,
+				}
```
stmcallister, line 536: This is causing an `undefined: extractor` error. You should be able to fix it by adding `var extractor := map[string]interface{}` just before the for loop. And then change these `extractor` assignments to use just `=`.

stmcallister (changes requested):
Thank you for making these changes, and welcome to Go! 👍 This project was my introduction to Go as well. I like what you're doing, but need to request one change that is mentioned later in the review. Thanks!

pagerduty/resource_pagerduty_service_integration.go
bohdanborovskyi-ma: I've pushed the fix and verified that provider compiles on my local machine.
Faced issues when tried to run acceptance test, so wasn't able to verify changes in it (yet it's very minor).

stmcallister:
Thank you for fixing that! Can you also vendor your branch (`go mod vendor`)? There are some conflicts due to a recent merge. Thanks!

bohdanborovskyi-ma:
> Thank you for fixing that! Can you also vendor your branch (`go mod vendor`)? There are some conflicts due to a recent merge. Thanks!

I've aligned my PR with master branch so there shouldn't be any conflicts (at least, from git perspective)

PR #483, opened 2022-04-18
Enforce 0 second time restriction on schedules

drastawi:
PagerDuty API currently zeros out any time, at least on the schedules endpoint, not at full minute resulting in a spurious diff and a 500 error. Best to ensure these values do not pass the plan.


stmcallister (approved):
👍 Thanks, @drastawi !!

PR #491, opened 2022-05-06
A pair of minor ACC testing improvements

metavida:
1. Add instructions to the README about running a small number of test
2. Switch tests to use `@foo.test` email addresses instead of `@foo.com` addresses. The goal of this change is to be certain that PagerDuty never sends a real email to a random domain when the acceptance tests run. This relies on the fact that [the .test TLD](https://en.wikipedia.org/wiki/.test) will never be publicly addressable.

I updated the CHANGELOG but marked the new version as unreleased because I don't think these changes warrant an actual release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.4.2 (Unreleased)
+
+IMPROVEMENTS:
+* Use "@foo.test" email addresses in tests
+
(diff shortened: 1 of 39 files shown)


stmcallister (approved):
Awesome! Thanks for making these changes. The acceptance tests all passed! 🎉

PR #494, opened 2022-05-10
[ORCA-3475] Allow deleting event orchestration team from an orchestration

alenapan:
### Changes
- Add support for "unsetting" the team
- Change Event Orchestration tests to have better coverage

### Testing
image

### DIFF
--- a/pagerduty/resource_pagerduty_event_orchestration.go
+++ b/pagerduty/resource_pagerduty_event_orchestration.go
@@ -89,7 +89,7 @@ func buildEventOrchestrationStruct(d *schema.ResourceData) *pagerduty.EventOrche
 	if attr, ok := d.GetOk("team"); ok {
 		orchestration.Team = expandOrchestrationTeam(attr)
 	} else {
-		var tId *string = nil
+		var tId *string
 		orchestration.Team = &pagerduty.EventOrchestrationObject{
 			ID: tId,
 		}
@@ -176,8 +176,19 @@ func resourcePagerDutyEventOrchestrationUpdate(d *schema.ResourceData, meta inte
 
 	log.Printf("[INFO] Updating PagerDuty Event Orchestration: %s", d.Id())
 
-	if _, _, err := client.EventOrchestrations.Update(d.Id(), orchestration); err != nil {
-		return err
+	retryErr := resource.Retry(10*time.Second, func() *resource.RetryError {
+		if _, _, err := client.EventOrchestrations.Update(d.Id(), orchestration); err != nil {
+			if isErrCode(err, 400) || isErrCode(err, 429) {
+				return resource.RetryableError(err)
+			}
+			return resource.NonRetryableError(err)
+		}
+
+		return nil
+	})
+
+	if retryErr != nil {
+		return retryErr
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


pagerduty/resource_pagerduty_event_orchestration.go
```diff
@@ -88,6 +88,11 @@ func buildEventOrchestrationStruct(d *schema.ResourceData) *pagerduty.EventOrche
 
 	if attr, ok := d.GetOk("team"); ok {
 		orchestration.Team = expandOrchestrationTeam(attr)
+	} else {
+		var tId *string = nil
+		orchestration.Team = &pagerduty.EventOrchestrationObject{
+			ID: tId,
+		}
```
alenapan, line 95: This is the main change - we are now explicitly setting the team ID to `nil` if it's not present on the resource

pagerduty/resource_pagerduty_event_orchestration.go
```diff
@@ -97,7 +102,7 @@ func expandOrchestrationTeam(v interface{}) *pagerduty.EventOrchestrationObject
 	var team *pagerduty.EventOrchestrationObject
 	t := v.([]interface{})[0].(map[string]interface{})
 	team = &pagerduty.EventOrchestrationObject{
-		ID: t["id"].(string),
+		ID: stringTypeToStringPtr(t["id"].(string)),
```
alenapan, line 105: Team ID is now a string pointer because, unlike strings, pointers can be assigned a `nil` value, so we need to convert it to string.

pagerduty/resource_pagerduty_event_orchestration_test.go
```diff
@@ -62,9 +62,16 @@ func TestAccPagerDutyEventOrchestration_Basic(t *testing.T) {
 			{
 				Config: testAccCheckPagerDutyEventOrchestrationConfigNameOnly(name),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckPagerDutyEventOrchestrationExists("pagerduty_event_orchestration.nameonly"),
+					testAccCheckPagerDutyEventOrchestrationExists("pagerduty_event_orchestration.foo"),
```
alenapan, line 65: I modified the tests a bit to cover the following flow:
- create an orchestration that only has `name`
- update the orchestration to also have `description` and `team`
- update `name`, `description`, `team`
- delete `description` and `team`

pagerduty/resource_pagerduty_event_orchestration.go
tyurina: Question because I'm a noob, why can't it just be `var tId *string` aren't string pointers by default nil?

alenapan: Very good question, let me test it and see if we can just declare it :)

alenapan: @tyurina I think you're right, we don't need to instantiate the variable. What I noticed is that Step 2 in the test sporadically fails with an "invalid team id" message., possibly because the team is created right before the orchestration. I think I may need to add 1 retry into the update method to fix it.

alenapan:
Ran the tests a couple of times after adding the retry logic and am not getting the "Invalid team ID" error anymore:
image

PR #495, opened 2022-05-11
Event orchestration unrouted resource

alexzakabluk:
Adds unrouted orchestration resource


pagerduty/resource_pagerduty_event_orchestration_path_unrouted_test.go
```diff
@@ -0,0 +1,477 @@
...
+			"pagerduty_escalation_policy",
+			"pagerduty_service",
+			"pagerduty_event_orchestration",
+		},
+	})
+}
+
+func TestAccPagerDutyEventOrchestrationPathUnrouted_Basic(t *testing.T) {
+	team := fmt.Sprintf("tf-name-%s", acctest.RandString(5))
+	escalation_policy := fmt.Sprintf("tf-%s", acctest.RandString(5))
```
alenapan, line 28: Can we rename `escalation_policy` to `escalationPolicy` to match Go naming conventions?

pagerduty/resource_pagerduty_event_orchestration_path_unrouted_test.go
```diff
@@ -0,0 +1,477 @@
...
+						"pagerduty_event_orchestration_unrouted.unrouted", "sets.0.id", "start"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_event_orchestration_unrouted.unrouted", "sets.0.rules.0.conditions.#", "2"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_event_orchestration_unrouted.unrouted", "sets.0.rules.0.actions.0.route_to", "child-1"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_event_orchestration_unrouted.unrouted", "sets.0.rules.0.conditions.0.expression", "event.severity matches part 'info'"),
+					resource.TestCheckResourceAttr(
+						"pagerduty_event_orchestration_unrouted.unrouted", "sets.0.rules.0.conditions.1.expression", "event.severity matches part 'warning'"),
+					resource.TestCheckResourceAttr(
```
alenapan, line 86: I took a slightly different path and extracted checking attributes for rule actions in a [separate function](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/ORCA-3465-orchestration-service-resource/pagerduty/resource_pagerduty_event_orchestration_path_service_test.go#L60-L79). But now that I see this approach it seems more readable and maybe we should keep all checks within the test step. I think I'm going to make changes on my branch.

pagerduty/resource_pagerduty_event_orchestration_path_unrouted_test.go
alexzakabluk: Oh, my bad🙂. Forgot about this one. Changing right now

alexzakabluk: Done

alenapan (approved):

PR #501, opened 2022-05-18
[ORCA-3467] Orchestration and path resource documentation

Pari-Dhanakoti:


website/docs/r/event_orchestration_resource.html.markdown
```diff
@@ -28,166 +23,244 @@ resource "pagerduty_event_orchestration" "my_monitor" {
...
+EventOrchestrations can be imported using the `id`, e.g.
+
+```
+$ terraform import pagerduty_event_orchestration.main 19acac92-027a-4ea0-b06c-bbf516519601
 ```
 
-**Argument Reference:** supports same rule, set, and catch-all arguments as the [Unrouted Orchestration endpoint](https://developer.pagerduty.com/api-reference/b3A6MzU3MDU0NDE-update-the-unrouted-orchestration-for-a-global-event-orchestration) of the Public API
+# pagerduty_event_orchestration_service
+
+When integrations exist on a service, [Service Orchestrations](https://support.pagerduty.com/docs/event-orchestration#service-orchestrations) can be used to evaluate incoming events against each of its rules, beginning with the rules in the "start" set. When a matching rule is found, it can modify and enhance the event and can route the event to another set of rules within this Service Orchestration for further processing. As a prerequisite, please ensure the service is set to evaluate orchestrations using the [Service Orchestration Active status endpoint](https://developer.pagerduty.com/api-reference/855659be83d9e-update-the-service-orchestration-active-status-for-a-service)
```
Pari-Dhanakoti, line 56: Frank confirmed that we do not want to Terraform the active status toggle for a service to evaluate through Orchestration instead of ruleset. Adding it as part of the description here, just so there is a mention of this setting somewhere.

website/docs/r/event_orchestration_path_resource.html.markdown
```diff
@@ -0,0 +1,222 @@
...
+* `catch_all`
+  * `actions`
+    * `suppress` - The suppress action for catch_all rule. This is always True.
+
+## Import
+
+Unrouted Orchestration path can be imported using the `id` of the event orchestration, e.g.
+
+```
+$ terraform import pagerduty_event_orchestration_unrouted 1b49abe7-26db-4439-a715-c6d883acfb3e
```
Pari-Dhanakoti, line 221: We are yet to test out imports. This command might require a change based on what we find during QA.

website/docs/r/event_orchestration_resource.html.markdown
```diff
@@ -28,166 +23,243 @@ resource "pagerduty_event_orchestration" "my_monitor" {
...
-    }
-  ]
-  # In this example the user has defined their own "all other unrouted events" rule with no conditions
-  # so they aren't defining global catch_all behavior for the pagerduty_event_orchestration_unrouted resource
-  # but using the catch_all would be a totally legit alternative.
-}
+The following attributes are exported:
+
+* `id` - The ID of the Event Orchestration.
+* `integrations` - Routing keys routed to this Event Orchestration.
```
metavida, line 44: **Question:** Won't `integrations` be an Array of objects? I wonder if we should somehow document the attributes available for each `integration`.

**Bonus:** Demonstrate how use the attribute to integrate with an event-provider like DataDog: https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/integration_pagerduty_service_object

metavida (commented):
I didn't get _all_ the way through, but wanted to post these comments before the end of my day. 

Comments listed as **Requested Resource Change** are changes we should discuss making to the resource itself. And if we **do** decide to change the resource then the API docs would need to be adjusted too.

website/docs/r/event_orchestration_path_resource.html.markdown
```diff
@@ -0,0 +1,220 @@
...
+  # This example assumes services used in the route_to configuration already exists. So it does not show creation of service resource.
+  # The first rule matches only if either of the two conditions are matched
+  # The second rule matches always
+  # The catch_all routes to unrouted path
+  # but using the catch_all to route to another service is allowed.
+  resource "pagerduty_event_orchestration_router" "router" {
+    parent {
+      # id of the Global Event Orchestrartion
+      id = pagerduty_event_orchestration.my_monitor.id
+    }
```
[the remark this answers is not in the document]
alenapan: I agree, it will be more readable if the team is a flat property. Regarding the parent object, it's an event_orchestration for router/unrouted and a service for a service path. We can have it configured as a `parent` or `event_orchestration` for router/unrouted and `service` for service paths

Pari-Dhanakoti:
> I didn't get _all_ the way through, but wanted to post these comments before the end of my day.
> 
> Comments listed as **Requested Resource Change** are changes we should discuss making to the resource itself. And if we **do** decide to change the resource then the API docs would need to be adjusted too.

All comments are addressed except for `catch_all` being optional as we weren't able to  generate default value for map/list. Team agrees to circle back on it if we have time to address it after all outstanding items + QA / GA without it as it is not a breaking change. I will update the internal terraform resource once the proposed resource changes (flattening parent and team blocks to be arguments) are tested. This documentation already reflects them

website/docs/r/event_orchestration.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The ID of the Event Orchestration.
+* `integrations` - List of integrations for the Event Orchestration.
+  * `id` - ID of the integration
+  * `parameters`
+    * `routing_key` - Routing key that routes to this Orchestration.
+    * `type` - Type of the routing key. `global` is the default type.
```
metavida, line 44: **Kudos:** Yay!! This level of detail is _super_ helpful!

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+    * `key` - (Required) Name to identify the header
+    * `value` - (Required) Value of this header
+  * `parameters` - (Optional) Specify custom key/value pairs that'll be included in the webhook request's JSON payload.
+    * `key` - (Required) Name to identify the parameter
+    * `value` - (Required) Value of this parameter
+* `severity` - (Optional) sets Severity of the resulting alert. Allowed values are: `info`, `error`, `warning`, `critical`
+* `event_action` - (Optional) sets whether the resulting alert status is trigger or resolve. Allowed values are: `trigger`, `resolve`
+* `variables` - (Optional) Populate variables from event payloads and use those variables in other event actions.
+  * `name` - (Required) The name of the variable
+  * `path` - (Required) Path to a field in an event, in dot-notation. This supports both PagerDuty Common Event Format [PD-CEF](https://support.pagerduty.com/docs/pd-cef) and non-CEF fields. Eg: Use `event.summary` for the `summary` CEF field. Use raw_event.fieldname to read from the original event `fieldname` data.
```
metavida, line 176: **Not a blocker:** I love that we're giving some examples of paths here. I also wonder if we could/should link to the PCL Path documentation here: https://developer.pagerduty.com/docs/ZG9jOjM1NTE0MDc0-pcl-overview#paths

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+  * `type` - (Required) Only `regex` is supported
+  * `value` - (Required) The Regex expression to match against. Must use valid [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) syntax.
+* `extractions` - (Optional) Replace any CEF field or Custom Details object field using custom variables.
+  * `template` - (Optional) A value that will be used to populate the `target` field. The configuration can include variables extracted from the payload by using string interpolation. Eg: If you have defined a variable called `hostname` you can set extraction `template` to `High CPU on variables.hostname server` to use the variable in extraction.  This field can be ignored for `regex` based replacements.
+  * `target` - (Required) The PagerDuty Common Event Format [PD-CEF](https://support.pagerduty.com/docs/pd-cef) field that will be set with the value from the `template` or based on `regex` and `source` fields.
```
metavida, line 181: **Not a blocker:** I'd definitely recommend `target` be moved to the top of the list because it's the only required field. This one is especially tricky to document, though, because `template` and `regex`+`source` are mutually exclusive. But having this sentence at the top is likely to help.

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+  * `name` - (Required) The name of the variable
+  * `path` - (Required) Path to a field in an event, in dot-notation. This supports both PagerDuty Common Event Format [PD-CEF](https://support.pagerduty.com/docs/pd-cef) and non-CEF fields. Eg: Use `event.summary` for the `summary` CEF field. Use raw_event.fieldname to read from the original event `fieldname` data.
+  * `type` - (Required) Only `regex` is supported
+  * `value` - (Required) The Regex expression to match against. Must use valid [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) syntax.
+* `extractions` - (Optional) Replace any CEF field or Custom Details object field using custom variables.
+  * `template` - (Optional) A value that will be used to populate the `target` field. The configuration can include variables extracted from the payload by using string interpolation. Eg: If you have defined a variable called `hostname` you can set extraction `template` to `High CPU on variables.hostname server` to use the variable in extraction.  This field can be ignored for `regex` based replacements.
```
metavida, line 180: **Blocker:** Missing curly braces within the template. In this case, I think re-using the wording from the UI's help modal might be nice!

Event Orchestration 2022-05-20 15-14-37


```suggestion
  * `template` - (Optional) A string that will be used to populate the `target` field. You can reference variables or event data within your template using double curly braces. For example:
     <br />Use variables named `ip` and `subnet` with a template like: `{{variables.ip}}/{{variables.subnet}}`
     <br />Combine the event severity & summary with template like: `{{event.severity}}: {{event.summary}}`
```

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+  * `target` - (Required) The PagerDuty Common Event Format [PD-CEF](https://support.pagerduty.com/docs/pd-cef) field that will be set with the value from the `template` or based on `regex` and `source` fields.
+  * `regex` - (Optional) The conditions that need to be met for the extraction to happen. Must use valid [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) syntax. This field can be ignored for `template` based replacements.
```
metavida, line 182: **Blocker:** I don't think this is quite right. Maybe we can re-use the wording from the API docs here?

```suggestion
  * `regex` - (Optional) A [RE2 regular expression](https://github.com/google/re2/wiki/Syntax) that will be matched against field specified via the `source` argument. If the regex contains one or more capture groups, their values will be extracted and appended together. If it contains no capture groups, the whole match is used. This field can be ignored for `template` based extractions.
```

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+  * `source` - (Optional) Field where the data is being copied from. Must be a PagerDuty Common Event Format [PD-CEF](https://support.pagerduty.com/docs/pd-cef) field. This field can be ignored for `template` based replacements.
```
metavida, line 183: **Blocker:** Again, trying to re-use the API doc wording here. Also seems worth linking to the PCL Path docs. Maybe something like this?

```suggestion
  * `source` - (Optional) The path to the event field where the `regex` will be applied to extract a value. You can use any valid [PCL path](https://developer.pagerduty.com/docs/ZG9jOjM1NTE0MDc0-pcl-overview#paths) like `event.summary` and you can reference previously-defined variables using a path like `variables.hostname`. This field can be ignored for `template` based extractions.
```

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+# pagerduty_event_orchestration_unrouted
+
+An Unrouted Orchestration allows users to create a set of Event Rules that will be evaluated against all events that don't match any rules in the Orchestration's Router.
+
+The Unrouted Orchestration evaluates events sent to it against each of its rules, beginning with the rules in the "start" set. When a matching rule is found, it can modify and enhance the event and can route the event to another set of rules within this Unrouted Orchestration for further processing.
+
+## Example of configuring a Unrouted Rules for an Orchestration
+
+  In this example the user has defined the unrouted orchestration with two rules, each routing to a different service.This example assumes services used in the route_to configuration already exists. So it does not show creation of service resource.As there is a single set, route_to is not defined for the rules.
+
```
metavida, line 18: **Blocker:** "two rules, each routing to a different service" I don't think it's possible to route to a service from the Unrouted Orchestration. Honestly, I think this whole sentence should go away.

```suggestion
```

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+  If there are more than one set, the rules in start set must define route_to with id of the next set
+  The first rule matches only if the condition is matched. The second rule matches always as there are no conditions.
```
metavida, line 20: **Not a Blocker:** Since this example doesn't demonstrate nested rules, I don't think we need to mention nested rules in this sentence specifically. Maybe something like this instead?

```suggestion
  In this example of an Unrouted Orchestration, first rule matches only if the condition is matched. The second rule matches all remaining events because it doesn't have any conditions.
```

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+
+  `catch_all` with empty `actions` block suppresses the alerts that do not match any rules. Using the `catch_all` to set severity, event_action, variables and extractions is also allowed.
```
metavida, line 22: **Blocker:** Again, I don't think suppress-by-default is correct (see my comment above).

Also, looking over this example (which is the same as [the API Docs example](https://developer.pagerduty.com/api-reference/953ddc1761403-update-the-unrouted-orchestration-for-a-global-event-orchestration)) makes me realize we have a bug! Seems like we should just use the `catch_all` block _instead_ of creating a rule with no conditions! In fact, that 2nd condition-less rule means the catch_all will **never** be used!

Might be worth changing this example to actually use the `catch_all` instead of 2 rules just to show how it can be used?

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+```
+## Argument Reference
+
+The following arguments are supported:
+
+* `service` - (Required) ID of the Service to which this Service Orchestration belongs to.
+* `sets` - (Required) A Service Orchestration must contain at least a "start" set, but can contain any number of additional sets that are routed to by other rules to form a directional graph.
+* `catch_all` - (Required) When none of the rules match an event, the event will be routed according to the catch_all settings.
```
metavida, line 139: **Blocker:** This is a Service Orchestration the `catch_all` cannot do routing. Maybe this instead?

```suggestion
* `catch_all` - (Required) the `catch_all` actions will be applied if an Event reaches the end of any set without matching any rules in that set.
```

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `event_orchestration` - (Required) The Event Orchestration to which this Unrouted Orchestration belongs to.
+* `sets` - (Required) An Unrouted Orchestration must contain at least a "start" set, but can contain any number of additional sets that are routed to by other rules to form a directional graph.
+* `catch_all` - (Required) When none of the rules match an event, the event will be routed according to the catch_all settings.
```
metavida, line 61: **Blocker:** Same as above, in an Unrouted Orchestration the `catch_all` cannot do routing. Maybe this instead?

```suggestion
* `catch_all` - (Required) the `catch_all` actions will be applied if an Event reaches the end of any set without matching any rules in that set.
```

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+
+### Sets (`sets`) supports the following:
+* `id` - (Required) The ID of this set of rules. Rules in other sets can route events into this set using the rule's `route_to` property.
+* `rules` - (Optional) The Unrouted Orchestration evaluates Events against these Rules, one at a time, and routes each Event based on the first rule that matches. If no rules are provided as part of Terraform configuration, the API returns empty list of rules.
```
metavida, line 65: **Blocker:** This isn't a router. Maybe this instead?

```suggestion
* `rules` - (Optional) The Unrouted Orchestration evaluates Events against these Rules, one at a time, and applies all the actions for first rule it finds where the event matches the rule's conditions. If no rules are provided as part of Terraform configuration, the API returns empty list of rules.
```

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+
+### Sets (`sets`) supports the following:
+* `id` - (Required) The ID of this set of rules. Rules in other sets can route events into this set using the rule's `route_to` property.
+* `rules` - (Optional) The service orchestration evaluates Events against these Rules, one at a time, and routes each Event based on the first rule that matches. If no rules are provided as part of Terraform configuration, the API returns empty list of rules.
```
metavida, line 143: **Blocker:** This isn't a Router so the phrase "routes each event" doesn't seem quite right to me. Maybe this instead?

* `rules` - (Optional) The Service Orchestration evaluates Events against these Rules, one at a time, and applies all the actions for first rule it finds where the event matches the rule's conditions. If no rules are provided as part of Terraform configuration, the API returns empty list of rules.

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+    rules {
+      label = "Update the summary of un-matched Critical alerts so they're easier to spot"
+      conditions {
+        expression = "event.severity matches 'critical'"
+      }
+      actions {
+        severity = "info"
+        extractions {
+            target = "event.summary"
+            template = "[Critical Unrouted] {{event.summary}}"
```
metavida, line 38: **Nit:** too deeply indented

```suggestion
          target = "event.summary"
          template = "[Critical Unrouted] {{event.summary}}"
```

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+* `label` - (Optional) A description of this rule's purpose.
+* `conditions` - (Optional) Each of these conditions is evaluated to check if an event matches this rule. The rule is considered a match if any of these conditions match. If none are provided, the event will `always` match against the rule.
+* `actions` - (Required) Actions that will be taken to change the resulting alert and incident, when an event matches this rule.
+* `disabled` - (Optional) Indicates whether the rule is disabled and would therefore not be evaluated.
+
+### Conditions (`conditions`) supports the following:
+* `expression`- (Required) A [PCL condition](https://developer.pagerduty.com/docs/ZG9jOjM1NTE0MDc0-pcl-overview) string.
+
+### Actions (`actions`) supports the following:
+* `route_to` - (Required) The ID of the target Service for the resulting alert.
```
metavida, line 77: **Blocker:** Nope, that's the router wording. Also it's Optional. Maybe this instead, taken from the REST API docs?

```suggestion
* `route_to` - (Optional) The ID of a Set from this Unrouted Orchestration whose rules you also want to use with event that match this rule.
```

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+* `label` - (Optional) A description of this rule's purpose.
+* `conditions` - (Optional) Each of these conditions is evaluated to check if an event matches this rule. The rule is considered a match if any of these conditions match. If none are provided, the event will `always` match against the rule.
+* `actions` - (Required) Actions that will be taken to change the resulting alert and incident, when an event matches this rule.
+* `disabled` - (Optional) Indicates whether the rule is disabled and would therefore not be evaluated.
+
+### Conditions (`conditions`) supports the following:
+* `expression`- (Required) A [PCL condition](https://developer.pagerduty.com/docs/ZG9jOjM1NTE0MDc0-pcl-overview) string.
+
+### Actions (`actions`) supports the following:
+* `route_to` - (Required) The ID of the target Service for the resulting alert.
```
metavida, line 155: **Blocker:** `route_to` is optional for service orchs & this text seems to be for the router. Maybe this wording taken from the API docs instead?

```suggestion
* `route_to` - (Optional) The ID of a Set from this Service Orchestration whose rules you also want to use with event that match this rule.
```

metavida (changes requested):
Phew! That's a _lot_ of documentation there! Just puts into scope the amount of functionality EO supports (and therefore our API & therefore the TF provider).

Let me know if you have questions about any of this!

website/docs/r/event_orchestration_unrouted.html.markdown
Pari-Dhanakoti: Good call out. It is true that for service path, the catch all behaviour leaves event as-is. But for `unrouted` the `catch all `always suppresses alert.
image
But I do agree that the example is confusing in the sense that 2nd rule will always match so catch_all will never be hit. I have removed the 2nd rule and moved it's alert action to the `catch_all`

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+
+## Example of configuring a Service Orchestration
+ This example shows creating `team`, `user`, `escalation policy` and `service` resources followed by the `service orchestration` resource. In this example the user has defined the service orchestration with nested rulesets.
+ The `start` set rule in this example routes to another set.
+ `catch_all` with empty `actions` block suppresses the alerts that do not match any rules. Other `actions` like `annotate`, `severity`, `priority`, `variables`, `extractions`, `webhooks`, `process automation` can be set in `catch_all` block.
```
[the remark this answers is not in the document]
Pari-Dhanakoti: As far the mention of paid feature, I tried to mention that the account requires Digital Operations plan and or top off as part of Business / Professional plan in order to add nested rulesets. I felt it was  making the comment section bloated. Also noticing that the [Event Intelligence](https://support.pagerduty.com/docs/event-intelligence#advanced-event-automation) KB page and [Orchestration](https://support.pagerduty.com/docs/event-orchestration) page does not mention about nested rulesets being a paid feature, I think it is better to leave it off of the  configuration comment section.

website/docs/r/event_orchestration_service.html.markdown
```diff
@@ -0,0 +1,215 @@
...
+  name                    = "My Web App"
+  auto_resolve_timeout    = 14400
+  acknowledgement_timeout = 600
+  escalation_policy       = pagerduty_escalation_policy.example.id
+  alert_creation          = "create_alerts_and_incidents"
+}
+
+data "pagerduty_priority" "p1" {
+  name = "P1"
+}
```
metavida, line 60: **Kudos:** Yay!! Makes this example super realistic _and_ copy/pastable for people wanting to learn more 🎉

website/docs/r/event_orchestration_unrouted.html.markdown
```diff
@@ -0,0 +1,103 @@
...
+
+An Unrouted Orchestration allows users to create a set of Event Rules that will be evaluated against all events that don't match any rules in the Orchestration's Router.
+
+The Unrouted Orchestration evaluates events sent to it against each of its rules, beginning with the rules in the "start" set. When a matching rule is found, it can modify and enhance the event and can route the event to another set of rules within this Unrouted Orchestration for further processing.
+
+## Example of configuring Unrouted Rules for an Orchestration
+
+In this example of an Unrouted Orchestration, the rule matches only if the condition is matched.
+Alerts created for events that do not match the rule will have severity level set to `info` as defined in `catch_all` block.
+`catch_all` for Unrouted Orchestration always suppressess the alert so it is not required to define the `suppress` action.
```
metavida, line 19: I'd argue it's more like "events that are sent to the Unrouted Orchestration will never belong to a service & therefore will **always** be suppressed, `catch_all` or not!" I'm going to remove this sentence for now, but feel free to add it (or a variation) back if you think it's removal makes things unclear. 👍 

```suggestion
```

metavida (approved):
I love where the documentation has landed! I'm planning to commit a pair of small tweaks to the branch, but then I think this PR is ready to merge 🚀

metavida (approved):
I love where the documentation has landed! I'm planning to commit a pair of small tweaks to the branch, but then I think this PR is ready to merge 🚀

Issue #400, opened 2021-10-04
Adding Response Plays integration to Services


sl-ediquas:
Hello,

In my team we have the need to associate response plays to services.  However it seems to be possible only by API and not by Terraform. Regarding API, here's the response from support : 

_Our engineers have confirmed that it is possible, but they have not yet documented it in our developer documentation.
Here are some examples. First when creating a Service:_

```
{
   "service": {
      "name": "Test API Service",
      "description": "Test Response Play",
      "escalation_policy": {
         "id": "PDXXXXX",
         "type": "escalation_policy_reference"
      },
      "response_play": {
         "id": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
         "type": "response_play_reference"
      }
   }
}
```
_Second, when updating a Service (/services/:id):_

```
{
   "service": {
      "response_play": {
         "id": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
         "type": "response_play_reference"
      }
   }
}
```
It would be great to have this feature reflected also in terraform resources, many thanks !

jjm:
The API docs seem to have been updated with details of assigning a response play to a service since this was raised, however I can't see this listed in the [CHANGELOG](https://developer.pagerduty.com/api-reference/ZG9jOjUxNzk5-changelog).

mattjones753:
I've put together a draft PR to implement this, just awaiting a PR in go-pagerduty first; https://github.com/heimweh/go-pagerduty/issues/91

mattjones753:
The PR should now be ready for review; https://github.com/PagerDuty/terraform-provider-pagerduty/pull/515

imjaroiswebdev:
Thanks to the contribution of @mattjones753 `response_play` reference in `pagerduty_service` it's rolled out and will be part of our next release 🎉 So I'm proceeding to close this Issue. **Thank you all** ✌🏽😁

PR #515, opened 2022-05-30
add response_play field to service

mattjones753:
This PR should allow users to attach a response_play to a service, which is now present in the API. should mean https://github.com/PagerDuty/terraform-provider-pagerduty/issues/400 can be closed

depends on a merge to the underlying pagerduty client; https://github.com/heimweh/go-pagerduty/pull/90


imjaroiswebdev:
Hi there @mattjones753 I want to thank you for doing the integration for `response_play` on `pagerduty_service`, I have tested and found that when the service was about to be written and the service the first time doesn't have a `response_play` assigned the provider was crash. Now that is solved, and I took the chance to add the test cases and the necessary documentation.

I also rebased your branch with the current `master` state to be consistent at the time of merging.

I want to thank you again for this great work @mattjones753 🎉

Tests results...
Screen Shot 2022-08-04 at 20 35 53

imjaroiswebdev (approved):

PR #517, opened 2022-06-01
Event orch resource singular names

alenapan:


pagerduty/resource_pagerduty_event_orchestration_path_unrouted.go
```diff
@@ -75,14 +75,14 @@ func resourcePagerDutyEventOrchestrationPathUnrouted() *schema.Resource {
...
-												"variables": {
+												"variable": {
 													Type:     schema.TypeList,
 													Optional: true,
 													Elem: &schema.Resource{
 														Schema: eventOrchestrationPathVariablesSchema,
 													},
 												},
-												"extractions": {
+												"extraction": {
```
alexzakabluk, line 85: I have checked ruleset example https://github.com/PagerDuty/terraform-provider-pagerduty/blob/master/pagerduty/resource_pagerduty_ruleset_rule.go#L246 and it looks like extractions kept as plural there. Same with `subconditions`. Is that overlooked in rulesets?

alenapan: I believe it was either overlooked or added before the naming convention was established because variables that were added later are singular: https://github.com/PagerDuty/terraform-provider-pagerduty/blob/master/pagerduty/resource_pagerduty_ruleset_rule.go#L285

alexzakabluk (approved):

stmcallister (approved):
Thanks @alenapan !! 👍 🎉 🌮

Issue #456, opened 2022-02-08
Unable to set "No Priority" vs "Any Priority" for priorities in pagerduty_slack_connection


PitcherIT:
Hi there,

It seems through the provider I am unable to set the priority on a slack connection to the “No priority” option (not “Any priority” option).  Through the REST API directly you can set/return this via `priorities: []` (for "No priority"), versus `priorities: null` (for "Any priority").
With Terraform I have tried omitting the priorities value, using `priorities = []`, `priorities = null` and `priorities = tolist([])` but this actually sets it to the "Any priorities" option.

### Terraform Version
Terraform v1.1.3
on darwin_amd64
+ provider registry.terraform.io/pagerduty/pagerduty v2.2.1

### Affected Resource(s)
pagerduty_slack_connection

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Expected Behavior
If setting `priorities = []` it should set the Priorities option to "No priority"
If setting `priorities = null` or omitting, it should set the Priorities option to "Any priority"
If setting `priorities = []` and resource exists and is set to "Any priority", it should see there is a change required, and apply changes it to "No priority".

### Actual Behavior
If setting `priorities = []` it sets the Priorities option to "Any priority"
If setting `priorities = []` and resource exists and is set to "Any priority", it see's there is no change to make

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
-

### References
-

imjaroiswebdev:
Hi there @PitcherIT 👋🏽 We addressed this Issue in the release [v2.5.1](https://github.com/PagerDuty/terraform-provider-pagerduty/releases/tag/v2.5.1). Try to give a look at the [updated documentation](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/slack_connection) for the new considerations to take in count.

PR #519, opened 2022-06-02
Addressing pagerduty_slack_connection unable to set "No Priority" vs "Any Priority" for priorities configuration

imjaroiswebdev:
Addressing #456 

After validating the behaviour of PagerDuty's API in regards of configuring the priorities for Slack Connection resource and comparing it with the current capabilities of our TF Provider, I could make sure that there ware no way to configure a `No Priority` priorities setting sticking to the interface of the PagerDuty's API. So it was necessary to add a more explicit way to achieve this configuration. As a result We end up making a little upgrade to the priorities attribute interface:

* `priorities` - (Optional) Allows you to filter events by priority. Needs to be an array of PagerDuty priority IDs. Available through [pagerduty_priority](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/data-sources/priority) data source.
  - When omitted or set to an empty array (`[]`) in the configuration for a Slack Connection, its default behaviour is to set `priorities` to `No Priority` value.
  - When set to `["*"]` its corresponding value for `priorities` in Slack Connection's configuration will be `Any Priority`.

Screen Shot 2022-06-09 at 00 30 11

Test results:

### DIFF
--- a/pagerduty/resource_pagerduty_slack_connection.go
+++ b/pagerduty/resource_pagerduty_slack_connection.go
@@ -12,6 +12,7 @@ import (
 )
 
 const AppBaseUrl = "https://app.pagerduty.com"
+const StarWildcardConfig = "*"
 
 func resourcePagerDutySlackConnection() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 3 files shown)


stmcallister (approved):
Very cool! Love the idea of adding the `*` syntax! Thank you!! 👍 🌮 🎉

mrwacky42:
Is it me, or is this a breaking change? It looks like the default value is the opposite of the old default value.

```
 resource "pagerduty_slack_connection" "slack" {
        id                = "<redacted>"
        # (7 unchanged attributes hidden)

!       config {
!           priorities = [
-               "*",
            ]
            # (1 unchanged attribute hidden)
        }
    }
```

Issue #520, opened 2022-06-06
Unable to switch off alert grouping on a service


pzhuravlenkopd:
### Terraform Version

terraform -v
Terraform v1.2.2
on darwin_amd64
+ provider registry.terraform.io/pagerduty/pagerduty v2.4.1

### Affected Resource(s)
- pagerduty service resource

Has the issue appeared to affect multiple resources? - tested only with PD service creations for now

### Terraform Configuration Files
```hcl

# Configure the PagerDuty provider
provider "pagerduty" {
  token = "${var.pagerduty_token}"
}

# Create PagerDuty team - Automation
resource "pagerduty_team" "automation" {
  name        = "Automation"
  description = "All automation engineers" 
}

resource "pagerduty_user" "dave" {
  name  = "Dave Bailey"
  email = "Dave.Bailey@pagerduty.demo"
  role = "user"
  time_zone = "Europe/London"

}
resource "pagerduty_user" "kate" {
  name  = "Kate Chapman"
  email = "Kate.Chapman@pagerduty.demo"
  role = "admin"
  time_zone = "Europe/London"
}

resource "pagerduty_team_membership" "teamDave" {
  user_id = "${pagerduty_user.dave.id}"
  team_id = "${pagerduty_team.automation.id}"
}
resource "pagerduty_team_membership" "teamkate" {
  user_id = "${pagerduty_user.kate.id}"
  team_id = "${pagerduty_team.automation.id}"
}


# Create PagerDuty Schedules
resource "pagerduty_schedule" "automation_sch" {
  name      = "Automation On-call Schedule"
  time_zone = "Europe/London"
  
  layer {
    name                         = "Layer 3 - Daily Rotation Dave"
    start                        = "2020-06-09T12:00:00-01:00"
    rotation_virtual_start       = "2020-06-09T12:00:00-01:00"
    rotation_turn_length_seconds = 86400
    users                        = ["${pagerduty_user.dave.id}"]
    restriction {
    type = "daily_restriction"
    start_time_of_day = "12:00:00"
    duration_seconds = 21600
    }
  }
  layer {
    name                         = "Layer 4 - Backup Daily Rotation Kate"
    start                        = "2020-06-09T18:00:00-01:00"
    rotation_virtual_start       = "2020-06-09T18:00:00-01:00"
    rotation_turn_length_seconds = 86400
    users                        = ["${pagerduty_user.kate.id}"]
    restriction {
    type = "daily_restriction"
[...]

PR #527, opened 2022-06-10
Address: Unable to switch off alert grouping on a service

imjaroiswebdev:
Addressing #520 

With this fix the Provider will not crash when trying to switch off alert grouping on a service, and also it will accept an interface more alike with the API and GUI.

Tests result:
Screen Shot 2022-06-10 at 18 44 36


stmcallister (approved):
Very nice! That util function will be useful! 👍 🌮 🎉

Issue #526, opened 2022-06-10
Error: The terraform-provider-pagerduty_v2.5.1 plugin crashed!


mikebryant:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

```
terraform -v
Terraform v0.15.0
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v4.17.1
+ provider registry.terraform.io/pagerduty/pagerduty v2.5.1

Your version of Terraform is out of date! The latest version
is 1.2.2. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_service_dependency

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

https://gist.github.com/mikebryant/e3c457f2f32d26ef0f82791b85ef6970

### Expected Behavior
What should have happened?

Deleted a dependency

### Actual Behavior
What actually happened?

Crashed

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

stmcallister:
Hi @mikebryant! Thanks for posting this issue. Unfortunately, I'm unable to reproduce the error. Can you provide more context into you what led to the crash? For example, would you be able to share the `pagerduty_service_dependency` definition in question?

mikebryant:
[...]

PR #530, opened 2022-06-16
add input validation and drift detection during deletion for `service_dependency`

imjaroiswebdev:
Address: #526 

Documentation improvement:

Dependency supporting and dependent service supports the following:

* `id` - (Required) The ID of the service dependency.
* `type` - (Required) Can be `business_service` or `service`.

Tests results...
Screen Shot 2022-06-16 at 16 57 45


stmcallister (approved):
Nice! I like it!! 👍 🎉 🌮

Issue #476, opened 2022-03-24
Terraform often fails with "http2: server sent GOAWAY and closed the connection"


foxylion:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

```
➜ terraform -v
Terraform v1.1.7
on linux_amd64
```

Provider: `pagerduty/pagerduty@v2.3.0`

### Affected Resource(s)

It affects all resources, or at least `pagerduty_escalation_policy` and `pagerduty_ruleset`.

### Terraform Configuration Files

I think this is independent from the resources, but we have a lot (hundreds) of the following resources/data:

- 285x data.pagerduty_ruleset
- 342x data.pagerduty_escalation_policy
- 150x resource.pagerduty_service
- 150x resource.pagerduty_ruleset_rule

### Debug Output

Due to privacy concerns I can't share the debug log output, but here is the errors we see in the normal log output:

- [http2: server sent GOAWAY and closed the connection; LastStreamID=199, ErrCode=NO_ERROR](https://gist.github.com/foxylion/faa69e3ed64e800c77cf1438fa14db65)
- [read tcp 10.1.0.38:56890->35.162.7.45:443: read: connection reset by peer](https://gist.github.com/foxylion/9b6621e5199a10142876b76cbcb73bc5)

### Panic Output

none

### Expected Behavior

I would expect no errors that indicate the PagerDuty API does not like the Terraform requests.

### Actual Behavior

I would just expect it works without any issues.

### Steps to Reproduce

1. Create a very large terraform project, with the above resources/data counts
3. Run `terraform plan`
4. Expect it to fail sometimes with the above mentioned errors.

alexs77:
Same here. We're not as big, though. We have just about 200x pagerduty_service and a few pagerduty_business_service.

PR #537, opened 2022-07-01
Remove 429 check on remaining data sources

cmrberry:
Similar to #521, this builds on #507 and removes the check for the 429 error code response and relax the `resource.Retry` to retry on all errors returned by the API for the remaining data sources that have not already been changed, since we're still getting reports similar to #476.


imjaroiswebdev:
Thank you so much! 👏🏽🎉

PR #543, opened 2022-07-13
update to go-version 1.17

stmcallister:
updating go version to be compatible with goreleaser.


imjaroiswebdev (approved):
Thank you Scott! 🌮

PR #548, opened 2022-07-21
provide instructions and make target for local install

jedelson-pagerduty:
IME it has been convenient to have an easy way to install a snapshot locally. I played with a few ways of doing this but since goreleaser is being used for the releases, it is the best way to create the snapshots (and then naturally use the goreleaser metadata to drive the actual local file move)


jedelson-pagerduty:
The instructions currently in the README work fine; no need for this complexity.

PR #552, opened 2022-07-28
use Provider.UserAgent to generate the user agent header

jedelson-pagerduty:
also updating goreleaser configuration to capture version and commit in a way that is easily accessible by the provider configure code


pagerduty/provider.go
```diff
@@ -97,13 +101,26 @@ func Provider() *schema.Provider {
 	}
 
 	p.ConfigureFunc = func(d *schema.ResourceData) (interface{}, error) {
-		terraformVersion := p.TerraformVersion
```
jedelson-pagerduty, line 100: Note that the *only* thing that was happening in `ConfigureFunc` was normalizing the Terraform version. That should not be necessary when using `Provider.UserAgent` so I just moved the body of `providerConfigure` here. I didn't see a reason to have a separate function.

.goreleaser.yml
```diff
@@ -5,7 +5,7 @@ builds:
   flags:
     - -trimpath
   ldflags:
-    - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
+    - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X github.com/terraform-providers/terraform-provider-pagerduty/pagerduty.version={{.Version}} -X github.com/terraform-providers/terraform-provider-pagerduty/pagerduty.commit={{.Commit}}'
```
jedelson-pagerduty, line 8: `main.version` and `main.commit` could be removed here -- those variables don't exist so the current (before this change) configuration is kind of a no-op.

Issue #534, opened 2022-06-23
Terraform unable to destroy and replace resources due to dependencies


elizajanus-ally:
### Terraform Version
v0.13.7

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_schedule
- pagerduty_escalation_policy
- pagerduty_team_membership

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
locals {

  example_user1 = {
    id = "EXAMPLE"
    role = "observer"
  }

  example_user2 = {
    id = "EXAMPLE"
    role = "observer"
  }

  example_members = [
    local.example_user1,
    local.example_user2
  ]
}

resource "pagerduty_team" "c3" {
  name        = "example"
  description = "example pagerduty team"
}

resource "pagerduty_team_membership" "example_membership" {
  for_each = { for example_members in local.example_members : example_members.id => example_members }
  user_id = each.key
  team_id = pagerduty_team.example.id
  role    = each.value.role
}

resource "pagerduty_schedule" "example_p1_schedule" {
  name      = "example p1 Primary Schedule"
  time_zone = "America/New_York"
  teams     = [pagerduty_team.example.id]
  
  layer {
    name                         = "example schedule"
    start                        = "2022-06-27T09:00:00.00Z"
    rotation_virtual_start       = "2022-06-27T09:00:00.00Z"
    rotation_turn_length_seconds = 86400*7
    users                        = local.example_members[*].id
  }
}

resource "pagerduty_schedule" "example_p1_schedule_secondary" {
  name      = "example p1 Secondary Schedule"
  time_zone = "America/New_York"
  teams     = [pagerduty_team.example.id]
  
  layer {
    name                         = "example schedule"
    start                        = "2022-07-04T09:00:00.00Z"
    rotation_virtual_start       = "2022-07-04T09:00:00.00Z"
    rotation_turn_length_seconds = 86400*7
    users                        = local.example_members[*].id
  }
}

resource "pagerduty_schedule" "example_p2_schedule" {
  name      = "example p2 Primary Schedule"
  time_zone = "America/New_York"
  teams     = [pagerduty_team.example.id]
  
  layer {
    name                         = "example schedule"
    start                        = "2022-06-27T09:00:00.00Z"
    rotation_virtual_start       = "2022-06-27T09:00:00.00Z"
    rotation_turn_length_seconds = 86400*7
[...]

Issue #427, opened 2021-12-02
pagerduty_escalation_policy should fail on plan if anyone on the policy is not a team member


drastawi:
It would be great to add some validation to the escalation policy resource. Currently the team members are not checked agains the target members and when these do not overlap, the escalation policy is invalid, yet the plan succeeds. 

### Terraform Version
2.1.1

### Affected Resource(s)
pagerduty_schedule

### Terraform Configuration Files
This already fails on the example form the official documentation if we remove the deprecated teams field. Side node, the doc

```
resource "pagerduty_team" "example" {
  name        = "Engineering"
  description = "All engineering"
}

resource "pagerduty_user" "example" {
  name  = "Earline Greenholt"
  email = "125.greenholt.earline@graham.name"
}

resource "pagerduty_escalation_policy" "example" {
  name      = "Engineering Escalation Policy"
  num_loops = 2
  teams     = [pagerduty_team.example.id]

  rule {
    escalation_delay_in_minutes = 10
    target {
      type = "user"
      id   = pagerduty_user.example.id
    }
  }
}
```

### Expected Behavior
The plan should fail early as the team membership is already known.

### Actual Behavior
The plan succeeds. Then fails on the apply as the user is not a team member.

### Steps to Reproduce
1. `terraform plan`

stmcallister:
@drastawi This didn't fail for me. 🤔

drastawi:
@stmcallister I did not clarify an important point which I did not realize was a prerequisite. The user had to have been a team member before to make this reproduce. So you should add the team membership code below: 

```
resource "pagerduty_team_membership" "example" {
  team_id = pagerduty_team.example.id
  user_id = pagerduty_user.example.id
}
```
and then delete it again.

So apply this first:
```
resource "pagerduty_team" "example" {
  name        = "Engineering"
  description = "All engineering"
}

resource "pagerduty_user" "example" {
  name  = "Earline Greenholt"
  email = "125.greenholt.earline@graham.name"
}

resource "pagerduty_team_membership" "example" {
  team_id = pagerduty_team.example.id
  user_id = pagerduty_user.example.id
}

resource "pagerduty_escalation_policy" "example" {
  name      = "Engineering Escalation Policy"
  num_loops = 2
  teams     = [pagerduty_team.example.id]

  rule {
    escalation_delay_in_minutes = 10
    target {
      type = "user_reference"
      id   = pagerduty_user.example.id
    }
  }
}
```
[...]

PR #558, opened 2022-08-17
Support for gracefully destroy `pagerduty_team_membership`

imjaroiswebdev:
Addresses #427 and #534 

This change is aimed to solve the recurring issue of getting an error of the kind...

```
Error: DELETE API call to https://api.pagerduty.com/teams/PEYO30I/users/PH5F8MH failed 400 Bad Request. Code: 2001, Errors: [User cannot be removed as they belong to an escalation policy on this team], Message: Invalid Input Provided
```

Without this update the API for dissociating an User from a Team currently can't remove it gracefully if the team is already linked to an Escalation Policy. So the solution is to dissociate the Team from the EP, remove the User from the Team and associate back the Team to the EP. Therefore, this change include the calls for following above mentioned flow.

> As direct consequence to this change deleting Users and Team Memberships linked with Escalation Policies will be solved too. 

## Test results...
Screen Shot 2022-08-17 at 18 16 01


stephanebruckert:
Using 2.9.3 and still getting `User cannot be removed as they belong to an escalation policy on this team`:

> module.first-last.pagerduty_team_membership.team_memberships["TEAM"]: Still destroying... [id=some-id, 1m50s elapsed]
> module.first-last.pagerduty_team_membership.team_memberships["TEAM"]: Still destroying... [id=some-id, 2m0s elapsed]
> 
> Error: DELETE API call to https://api.pagerduty.com/teams/team-id/users/user-id failed 400 Bad Request. Code: 2001, Errors: [User cannot be removed as they belong to an escalation policy on this team], Message: Invalid Input Provided

any idea @imjaroiswebdev?

deniojunior-hotmart:
Using `2.11.0` and I'm still getting this error

bschaeffer:
Same as @deniojunior-hotmart. @imjaroiswebdev is there a way we need to setup our resources so that these things are being deleted correctly? I tried setting a it such that the team_membership `depends_on` schedule but TF still tries to delete the membership first.

deck1187hw:
This is still an issue on `3.18.3` Any support would be appreciated. Really tried everything.

Issue #358, opened 2021-07-20
Cannot delete team with existing Schedules association(s)


m1n9o:
### Terraform Version
0.15.5

### Affected Resource(s)
Please list the resources as a list, for example:
- team
- schedule

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

The problem is much straightforward. I created a team with escalation policy and services.
I removed the escalation policy and services with terraform apply. After that, when I  try to remove the team.
It says cannot delete team with existing schedules association(s). It's quite annoying if we need to disassociate them manually.

### Panic Output
```
[2021-07-20T08:26:25Z] Error: DELETE API call to https://api.pagerduty.com/teams/P2WMR52 failed 400 Bad Request. Code: 6006, Errors: [Cannot delete team with existing Schedules association(s)], Message: Team has existing associations.
```
### Expected Behavior
It should be able to remove the team.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

NargiT:
it's limitation on pager duty. Since the schedule is linked to the team, you cannot delete the team. you need  to first delete the schedule then the team.

m1n9o:
> it's limitation on pager duty. Since the schedule is linked to the team, you cannot delete the team. you need to first delete the schedule then the team.
> 
> maybe you can used `depends_on` feature so terraform will first delete that perticular schedule before the team.

Why not just disassociate them so that terraform can destroy it. And keep the schedule at the same time.
`Schedule` should not be deleted and re-created that often.

edobry:
This is also an issue for me, I've tried configuring the `depends_on` in a variety of ways but couldn't get it to work. Ended up having to delete manually through the UI; would be great to be able to delete through TF!

NargiT:
What is funny is that you can create a schedule only if a responder is assign to it. but if you delete the user, the schedule still exists and a "shadow user take it's place". Feels like core design issues IMO. This make terraform a bit more tricky to use.

imjaroiswebdev:
@m1n9o @NargiT @edobry An update for fixing this issue was merged on #561 and It has been released in version **v2.6.1** of PagerDuty TF Provider, so updating to the new version hopefully You all would be able to gracefully delete Schedules without getting stuck with into this error. Please if You find any further issues related with this error after updating don't hesitate on reopening this Issue ✌🏽

drastawi:
[...]

Issue #337, opened 2021-05-20
Module doesn't understand the dependency between schedules and escalation policies


noqcks:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.12.26
+ provider.pagerduty v1.9.6

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_schedule
- pagerduty_escalation_policy

### Terraform Configuration Files
```hcl
resource "pagerduty_schedule" "example" {
  name      = "Daily Engineering Rotation"
  ...
}

resource "pagerduty_escalation_policy" "example" {
  name      = "Engineering Escalation Policy"
  teams     = [pagerduty_team.example.id]

    rule {
    escalation_delay_in_minutes = 10

    target {
      type = "schedule_reference"
      id   = pagerduty_schedule.example.id
    }
  }
}
```

### Scenario

The above `pagerduty_schedule` and `pagerduty_escalation_policy` exist in Pagerduty and TF state. I remove the code for the escalation policy _and_ the schedule and run `tf apply`.

### Expected Behavior

I expect the pagerduty module to understand the dependency between the schedule and the escalation policy and to delete the escalation policy first and the ~policy~ schedule second.

### Actual Behavior
```
Error: DELETE API call to https://api.pagerduty.com/schedules/TRFY575 failed 400 Bad Request. Code: 2001, Errors: [Schedule can't be deleted if it's being used by escalation policies], Message: Invalid Input Provided
Error: DELETE API call to https://api.pagerduty.com/schedules/TR2IAZ9 failed 400 Bad Request. Code: 2001, Errors: [Schedule can't be deleted if it's being used by escalation policies], Message: Invalid Input Provided
```

### Steps to Reproduce
1. Create a schedule linked to an escalation policy. 
2. Try delete both at the same time with `tf apply`

jashugan:
> I expect the pagerduty module to understand the dependency between the schedule and the escalation policy and to delete the escalation policy first and the policy second.

@noqcks -- I think you mean "escalation policy first and the schedule second".

matan129:
The same happens for Escalation Policies and Services.
If I have a Service S1 and Escalation Policy EP1, which is set to be the EP of S1, and then try to delete both resources, the provider will try to delete the Service first.

matan129:
[...]

PR #561, opened 2022-08-23
Support for gracefully destroy `pagerduty_schedule`

imjaroiswebdev:
Addresses #337, #358 and #534 

This change is aimed to solve the recurring issue when trying to destroy a `pagerduty_schedule` resource which ends with the following error log...
```
Error: DELETE API call to https://api.pagerduty.com/schedules/TRFY575 failed 400 Bad Request. Code: 2001, Errors: [Schedule can't be deleted if it's being used by escalation policies], Message: Invalid Input Provided
```

Currently the `pagerduty_schedule` destroy implementation can't surpass the above mentioned error, because when a Schedule is being used by one or many Escalation Policies, it's needed to remove the `schedule_reference` targets from the Escalation Policies, and only after that is completed a graceful delete operation can be executed on the Schedule. So, this change includes the API calls for following previously explained flow.

## Test Results...
Screen Shot 2022-08-23 at 18 29 40

### DIFF
--- a/pagerduty/resource_pagerduty_schedule.go
+++ b/pagerduty/resource_pagerduty_schedule.go
@@ -365,8 +365,23 @@ func resourcePagerDutyScheduleDelete(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
-	log.Printf("[INFO] Deleting PagerDuty schedule: %s", d.Id())
+	log.Printf("[INFO] Starting deletion process of Schedule %s", d.Id())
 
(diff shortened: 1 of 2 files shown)


stmcallister (approved):
Great implementation for this issue! Nice work, @imjaroiswebdev!! 🌮 🎉 🍕

nthienan:
when will this be released?

imjaroiswebdev:
> when will this be released?

Hi @nthienan This fix was made available in **v2.6.1** of **PagerDuty** Terraform Provider.

Issue #562, opened 2022-08-29
pagerduty_tag_assignment misconfiguration in the docs


Coolomina:
Bad resource configuration on [pagerduty_tag_assignment](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/tag_assignment) resource documentation, where the `pagerduty_tag` resource turns up with an unknown configuration option called `type`, which produces a `An argument named "type" is not expected here.` error when running `terraform validate` on any version of Terraform.

Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v1.2.8
- provider registry.terraform.io/pagerduty/pagerduty v2.6.1

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_tag_assignment

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "pagerduty_tag" "example" {
  label = "API"
  type  = "tag"
}
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform validate`

PR #563, opened 2022-08-29
Adds proper documentation for pagerduty_tag resource in tag_assignmen…

Coolomina:
Addresses #562

### DIFF
--- a/website/docs/r/tag_assignment.html.markdown
+++ b/website/docs/r/tag_assignment.html.markdown
@@ -15,7 +15,6 @@ A [tag](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODEwMA-assign-tags
 ```hcl
 resource "pagerduty_tag" "example" {
   label = "API"
-  type  = "tag"
 }
 resource "pagerduty_team" "engteam" {
   name = "Engineering"


johncoleman83:
Hi @imjaroiswebdev, I have reviewed this doc change, and looks good to me.

To provide some context, [The PagerDuty terraform docs](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/tag_assignment) for `pagerduty_tag_assignment` recommend putting the `type` as a supported argument as @Coolomina mentions clearly in the linked issue; however the `pagerduty_tag` resource does not support `type` as an argument [(see docs here)](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/tag). I also notice that none of the PagerDuty terraform resources (which is not the same as resources in the [REST API](https://developer.pagerduty.com/api-reference/3fe68a932a717-get-tags-for-entities) that do support the `type` attribute as a specifier of the resource itself) support the `type` argument unless it is used to distinguish a variety of the resource rather than the resource itself.

With that said, I think these changes should be added. I suppose I should clarify are the docs at `/website/docs` the proper docs for PagerDuty resource that we publish [here](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs)?

johncoleman83 (approved):

imjaroiswebdev:
@johncoleman83 You totally right on what you say and @Coolomina thank you so much for your contribution, this update will definitely help other users of the Provider 💪🏽 😎

PR #564, opened 2022-09-01
Test rule removal rather than update for Escalation Policy Dependent Schedule

drastawi:
Looks like there is a case missing in #561 

The code should work for rule update as well as rule removal.


drastawi:
@imjaroiswebdev PTAL

imjaroiswebdev (approved):
Hey @drastawi! One more time thank you for contributing with this patch that solves this missing case. I only rebased your commits to the current `master` state and refactor your patch just for code readability purpose.

Issue #550, opened 2022-07-22
"Intelligent" alert grouping type on pagerduty_service seems to require empty config block


DanielWeintraub:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
terraform 0.14.11

### Affected Resource(s)
- pagerduty_service

### Terraform Configuration Files
within a pagerduty_service block
```hcl
  alert_grouping_parameters {
    type = "intelligent"
    config {
      fields  = null
      timeout = 0
    }
  }
```

this used to work, but now returns:
```
module.monitoring.module.external_secrets.pagerduty_service.service: Creating...

Error: POST API call to https://api.pagerduty.com/services failed 400 Bad Request. Code: 2001, Errors: [Config is invalid.], Message: Invalid Input Provided

  on modules/pd_service/main.tf line 20, in resource "pagerduty_service" "service":
  20: resource "pagerduty_service" "service" {
```

It works fine if you remove the config block.

[The documentation](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service) says: 

> If type is set to intelligent or empty then config can be empty.

BillSidney:
Without the config block, terrafrom apply updates each existing service in-place  adding the config block

```
# pagerduty_service.this will be updated in-place
  ~ resource "pagerduty_service" "this" {
        id                      = "XXXXXX"
        name                    = "ServiceDaemon"
        # (11 unchanged attributes hidden)
      ~ alert_grouping_parameters {
            # (1 unchanged attribute hidden)
          - config {
              - fields  = [] -> null
              - timeout = 0 -> null
            }
        }
        # (1 unchanged block hidden)
    }
```

BillSidney:
More Details  - When attempting to change the description, updated in-place, using an empty config block fails.

```
# module.pagerduty_resource["XXXi"].pagerduty_service.this will be updated in-place
  ~ resource "pagerduty_service" "this" {
      ~ description             = <<-EOT
          ...
          - criticality: 0
          + criticality: 2

          - resiliency : 0
          + resiliency : 5
          ...
            
            Managed by Terraform
        EOT
        id                      = "XXXX"
        name                    = "xxxxxxx"
        # (11 unchanged attributes hidden)
        # (2 unchanged blocks hidden)
    }

[...]

PR #570, opened 2022-09-15
Address: `pagerduty_service.alert_grouping_parameters.config` block parsing

imjaroiswebdev:
Address #550 

This update is aimed to correctly handle the parsing for `pagerduty_service.alert_grouping_paramerts.config` block.

Test results...
Screen Shot 2022-09-15 at 17 29 16

### DIFF
--- a/pagerduty/resource_pagerduty_service.go
+++ b/pagerduty/resource_pagerduty_service.go
@@ -566,13 +566,21 @@ func expandAlertGroupingParameters(v interface{}) *pagerduty.AlertGroupingParame
 		return nil
 	}
 	riur := pre.(map[string]interface{})
-	if len(riur["type"].(string)) > 0 {
-		gt := riur["type"].(string)
-		alertGroupingParameters.Type = &gt
+	groupingType := ""
+	if riur["type"].(string) != "" {
+		groupingType = riur["type"].(string)
+		alertGroupingParameters.Type = &groupingType
 	}
 
-	if val, ok := riur["config"]; ok {
-		alertGroupingParameters.Config = expandAlertGroupingConfig(val)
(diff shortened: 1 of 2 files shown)


stmcallister (approved):
Looks good to me! Are there any changes that need to be made to the documentation because of this change?

imjaroiswebdev:
It's great that you point this out @stmcallister, because the issue being addressed by this PR is happening because the current implementation doesn't stick to the [documented API](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service#alert_grouping_parameters) when it has to do with updating `pagerduty_service.alert_grouping_parameters` configurations, and now with this update the Provider will able to consistently **create** and **update** `pagerduty_service.alert_grouping_parameters` resource configurations.

davidpellcb:
Hi all, found this PR through various googling. We're currently way behind on our provider version (2.1.1) but seeing constant unnecessary diffs in our plans that look like this for every single service that uses `alerting_group_parameters` with `type = "intelligent"`:

image

However, if we actually use `config: []` in the `alerting_group_parameters` we get an error:

image

I believe this is because `config` isn't meant to be used with `type = "intelligent"` based on the [docs](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service#alert_grouping_parameters):

image

Not sure if this is an issue with the PD API or the TF provider. Also unsure from reading the PR description if this will address our issue. Do I need to open a separate issue?

imjaroiswebdev:
Hi @davidpellcb the issue you are facing with your config is more related to the syntax, because your are trying to declare an Attribute named **config**, instead of a **config** block.

PR #573, opened 2022-09-22
resource_pagerduty_service: skip response_play with "null" value

Padorax:
Similar to fields "auto_resolve_timeout" and "acknowledge_timeout" etc, this allows to skip response_play by setting value to "null" besides not writing "response_play" field. This is easier for creating multiple services with modules when some have response_play while others don't.

### DIFF
--- a/pagerduty/resource_pagerduty_service.go
+++ b/pagerduty/resource_pagerduty_service.go
@@ -388,9 +388,11 @@ func buildServiceStruct(d *schema.ResourceData) (*pagerduty.Service, error) {
 	}
 
 	if attr, ok := d.GetOk("response_play"); ok {
-		service.ResponsePlay = &pagerduty.ResponsePlayReference{
-			ID:   attr.(string),
-			Type: "response_play_reference",
+		if attr.(string) != "null" {
+			service.ResponsePlay = &pagerduty.ResponsePlayReference{
+				ID:   attr.(string),
+				Type: "response_play_reference",
+			}
 		}
 	}
 	return &service, nil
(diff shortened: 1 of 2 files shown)


drastawi (approved):
Looks great. @imjaroiswebdev any chance we release this soon?

imjaroiswebdev (approved):
Patch and tests looks great! Thank you @Padorax and @drastawi. I'm proceeding to merge 💪🏽 ✌🏽

s3265176:
How Do I unset response_play on resource_pagerduty_service then?  Need to remove a response_play on a service.

Padorax:
@s3265176 Before this change, ```response_play=null``` will attempt to set the response_play id to string "null", which is not a valid value.
As described in the commit message, this change ensures "null" is accepted as an empty value (i.e., no response_play), which makes a consistent behaviour with many other terraform resources, and easier to use with module.

To unset response_play from a service, you can use ```response_play=null``` or ```respon_play=""```, both seems to works for me. Can you try and see if you have any issue with the solution?

s3265176:
@Padorax  I am on 2.10.2
My pagerduty_service is used with a for_each loop
setting `response_play = null`   `response_play = ""`   `response_play = " "`   `response_play = "null"` all failed to remove the 
response play from services. 

During terraform plan it did pick up the change 
` ~ response_play           = "17420b15-2702-4303-75b7-43e9f5fbdb53" -> "null"`

Padorax:
@s3265176 If the change is detected by "plan", but not actually applied, then seems like it's more broken than we fixed.

bjor-joh:
We are experiencing the same behaviour.

imjaroiswebdev:
Hey @s3265176 @Padorax @bjor-joh This issue should be solved with #680

PR #581, opened 2022-10-21
feat: add pagerduty_event_orchestrations datasource

fgouteroux:
Hello,

I want to add a new datasource pagerduty_event_orchestrations which allow to search multiples event_orchestration with a regex.

Example:
```
data "pagerduty_event_orchestrations" "example" {
  search = ".*Orchestration$"
}
```
I added:

- [x] doc
- [x] tests

### DIFF
--- a/pagerduty/data_source_pagerduty_event_orchestrations.go
+++ b/pagerduty/data_source_pagerduty_event_orchestrations.go
@@ -85,7 +85,10 @@ func dataSourcePagerDutyEventOrchestrationsRead(d *schema.ResourceData, meta int
 			return resource.RetryableError(err)
 		}
 
-		re := regexp.MustCompile(searchName)
+		re, err := regexp.Compile(searchName)
+		if err != nil {
+			return resource.NonRetryableError(fmt.Errorf("invalid regexp for name_filter provided %s", searchName))
+		}
 		for _, orchestration := range resp.Orchestrations {
 			if re.MatchString(orchestration.Name) {
 				eoList = append(eoList, orchestration)
--- a/pagerduty/data_source_pagerduty_event_orchestrations.go
+++ b/pagerduty/data_source_pagerduty_event_orchestrations.go
@@ -1,0 +1,127 @@
+package pagerduty
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func dataSourcePagerDutyEventOrchestrations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePagerDutyEventOrchestrationsRead,
+
+		Schema: map[string]*schema.Schema{
+			"search": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"event_orchestrations": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"integration": {
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"id": {
+										Type:     schema.TypeString,
+										Computed: true,
+									},
+									"parameters": {
+										Type:     schema.TypeList,
+										Computed: true,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"routing_key": {
+													Type:     schema.TypeString,
+													Computed: true,
+												},
+												"type": {
+													Type:     schema.TypeString,
+													Computed: true,
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourcePagerDutyEventOrchestrationsRead(d *schema.ResourceData, meta interface{}) error {
+	client, err := meta.(*Config).Client()
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] Reading PagerDuty Event Orchestrations")
+
+	searchName := d.Get("search").(string)
+
+	return resource.Retry(5*time.Minute, func() *resource.RetryError {
+		resp, _, err := client.EventOrchestrations.List()
+		if err != nil {
+			return resource.RetryableError(err)
+		}
+
+		var orchestrations []*pagerduty.EventOrchestration
+		re := regexp.MustCompile(searchName)
+		for _, orchestration := range resp.Orchestrations {
+			if re.MatchString(orchestration.Name) {
+				// Get orchestration matched by ID so we can set the integrations property
+				// since the list endpoint does not return it
+				orch, _, err := client.EventOrchestrations.Get(orchestration.ID)
+				if err != nil {
+					return resource.RetryableError(err)
+				}
+				orchestrations = append(orchestrations, orch)
+			}
+		}
+
+		if len(orchestrations) == 0 {
+			return resource.NonRetryableError(
+				fmt.Errorf("Unable to locate any Event Orchestration with the name: %s", searchName),
+			)
+		}
+
+		d.SetId(resource.UniqueId())
+		d.Set("search", searchName)
+		d.Set("event_orchestrations", flattenPagerDutyEventOrchestrations(orchestrations))
+
+		return nil
+	})
+}
+
+func flattenPagerDutyEventOrchestrations(orchestrations []*pagerduty.EventOrchestration) []interface{} {
+	var result []interface{}
+
+	for _, o := range orchestrations {
+		orchestration := map[string]interface{}{
+			"id":          o.ID,
+			"name":        o.Name,
+			"integration": flattenEventOrchestrationIntegrations(o.Integrations),
+		}
+		result = append(result, orchestration)
+	}
+	return result
+}
--- a/pagerduty/data_source_pagerduty_event_orchestrations.go
+++ b/pagerduty/data_source_pagerduty_event_orchestrations.go
@@ -16,7 +16,7 @@ func dataSourcePagerDutyEventOrchestrations() *schema.Resource {
 		Read: dataSourcePagerDutyEventOrchestrationsRead,
 
 		Schema: map[string]*schema.Schema{
-			"search": {
+			"name_filter": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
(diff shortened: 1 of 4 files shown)


fgouteroux:
any updates ?

gsreynolds:
Rebased to resolve conflicts; @imjaroiswebdev can you please review when able? Much appreciated

gsreynolds:
Hi @fgouteroux - just for context, it would help if you could elaborate on the use case here. 

From your example in the doc:
```hcl
resource "pagerduty_event_orchestration" "tf_orch_a" {
  name = "Test Event A Orchestration"
}
resource "pagerduty_event_orchestration" "tf_orch_b" {
  name = "Test Event B Orchestration"
}
data "pagerduty_event_orchestrations" "tf_my_monitor" {
  search = ".*Orchestration$"
}
```

I'm presuming in your scenario that you might not know ahead of time that EOs (A & B) exist that might match `".*Orchestration$"`, however... what are you then looking to do with the returned data? Thanks

fgouteroux:
> Hi @fgouteroux - just for context, it would help if you could elaborate on the use case here.
> 
> From your example in the doc:
> 
> ```terraform
> resource "pagerduty_event_orchestration" "tf_orch_a" {
>   name = "Test Event A Orchestration"
> }
> resource "pagerduty_event_orchestration" "tf_orch_b" {
>   name = "Test Event B Orchestration"
> }
> data "pagerduty_event_orchestrations" "tf_my_monitor" {
>   search = ".*Orchestration$"
> }
> ```
> 
> I'm presuming in your scenario that you might not know ahead of time that EOs (A & B) exist that might match `".*Orchestration$"`, however... what are you then looking to do with the returned data? Thanks

Hello @gsreynolds , We need to get the integration keys in order to configure our alertmanager. We have a naming convention for the event orchestration (the owner team name in the name of EO).

Today we have declare n datasources to get the event orchestration matching the exact name. Each datasource made a Pagerduty http call and it is useless to perform n call when we can do it once. 

At the end we need want to configure multiple alertmanager pagerduty_receivers, one for each integration_key.

```
data "pagerduty_event_orchestration" "messaging" {
  name = "myteam.messaging"
}

data "pagerduty_event_orchestration" "monitoring" {
  name = "myteam.monitoring"
}

module "mimir_alertmanager_config" {
  source             = "git::ssh://git@gitlab.example.com/terraform/modules/prometheus_alertmanager_config.git"
  default_route = {
    receiver = "pagerduty_default"
  }
  routes = [
    {
      receiver = "pagerduty_messaging"
      matchers = ["ecosystem=\"stack_messaging\""]
    },
    {
      receiver = "pagerduty_monitoring"
      matchers = ["ecosystem=\"stack_monitoring\""]
    }
  ]
  receivers_pagerduty = [
 
    {
      name        = "pagerduty_messaging"
      routing_key = "${data.pagerduty_event_orchestration.messaging.routing_keys[0]}"
    },
    {
      name        = "pagerduty_monitoring"
      routing_key = "${data.pagerduty_event_orchestration.monitoring.routing_keys[0]}"
    }
  ]
}
```

pagerduty/data_source_pagerduty_event_orchestrations.go
```diff
@@ -0,0 +1,127 @@
...
+	searchName := d.Get("search").(string)
+
+	return resource.Retry(5*time.Minute, func() *resource.RetryError {
+		resp, _, err := client.EventOrchestrations.List()
+		if err != nil {
+			return resource.RetryableError(err)
+		}
+
+		var orchestrations []*pagerduty.EventOrchestration
+		re := regexp.MustCompile(searchName)
```
imjaroiswebdev, line 88: Could you please change this by `regexp.Compile` and handle the error caused by invalid regex expression?

Additionally, please cover the regex compilation error handling with a test case.

pagerduty/data_source_pagerduty_event_orchestrations.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] Reading PagerDuty Event Orchestrations")
+
+	searchName := d.Get("search").(string)
+
+	return resource.Retry(5*time.Minute, func() *resource.RetryError {
+		resp, _, err := client.EventOrchestrations.List()
```
imjaroiswebdev, line 82: Please let's separate this `List` call retries from the `Get` retries. We shouldn't be doing nested retries.

website/docs/d/event_orchestrations.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+layout: "pagerduty"
+page_title: "PagerDuty: pagerduty_event_orchestrations"
+sidebar_current: "docs-pagerduty-datasource-event-orchestrations"
+description: |-
+  Get information about Global Event Orchestrations that you have created.
+---
+
+# pagerduty\_event_orchestrations
+
+Use this data source to get informations about specific Global [Event Orchestrations][1]
```
imjaroiswebdev, line 11: I suggest you the following wording...

> Use this data source to get information as a list about specific Global [Event Orchestrations][1] filtered by a Regular Expression provided.

imjaroiswebdev (changes requested):
Hi @fgouteroux I really appreciate your contribution and I'm also able to understand the value in this new data source you are introducing, however, I would ask you to first address a few comments I left as my review to proceed with its approval and merge. Thanks again! ✌🏽 

Please don't hesitate in asking for help or if you need me to elaborate more on my comments. I'm also open to push the patches to cover my comments in order to accelerate the merge of this PR. You only have to let me know 💪🏽

pagerduty/data_source_pagerduty_event_orchestrations.go
fgouteroux: @imjaroiswebdev I'm not sure to understand as this is the same code than pagerduty/data_source_pagerduty_event_orchestration.go.

Could you give to me an example ?

pagerduty/data_source_pagerduty_event_orchestrations.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/heimweh/go-pagerduty/pagerduty"
+)
+
+func dataSourcePagerDutyEventOrchestrations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePagerDutyEventOrchestrationsRead,
+
+		Schema: map[string]*schema.Schema{
+			"search": {
```
imjaroiswebdev, line 19: I suggest you change this attribute for `name_filter`, just to be more alined with common practices used by Hashicorp implementation in others Provider like the AWS one, which uses the attribute `filter` on listed data source to provide an interface for supplying search parameters, but in their case `filter` is a key-value structure, whereas in our case We are using exclusively the **name** as the lookup field.

gsreynolds (changes requested):
@fgouteroux Thanks for your use case and I understand better now

In my personal opinion, it seems not to be entirely in keeping with the declarative nature of Terraform. Still, I appreciate the scenario and acknowledge that “list” style resources exist elsewhere.

As @imjaroiswebdev has suggested, we would prefer to rename the attribute `search` to `name_filter`. If you look at similar “list” style data sources elsewhere they have a filter block for key/value pairs, but we only have a name (maybe team) here… For example:

[registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/availability_zones](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/availability_zones)
[registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/instances](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/instances)

pagerduty/data_source_pagerduty_event_orchestrations.go
imjaroiswebdev: Even though this code is similar to the one of single EO Data Source, what I mean with We shouldn't be nesting retries is that, because the code is triggering a retry inside a loop that is being scoped outside the loop, all the already done calls that probably resulted **ok** would be re executed, and that's why We should in this case scope the retry to each iteration of the loop, to avoid the abuse of the API calls. You can get inspiration from https://github.com/PagerDuty/terraform-provider-pagerduty/blob/5431f120ba880507d6069970f4f50e8391f2b568/pagerduty/resource_pagerduty_event_orchestration_path_unrouted.go#L215 on how to scope the retries inside each loop iteration. @fgouteroux again, please don't hesitate on asking for more help, to be more elaborated or even if you need it, I can directly push a commit with the refactor to solve this particular comment. I'll be happy to do it ✌🏽

fgouteroux: if you offer to do it, I accept. Thanks

imjaroiswebdev (approved):
All comments have been addressed ✌🏽

website/docs/d/event_orchestrations.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+resource "pagerduty_event_orchestration" "tf_orch_a" {
+  name = "Test Event A Orchestration"
+}
+
+resource "pagerduty_event_orchestration" "tf_orch_b" {
+  name = "Test Event B Orchestration"
+}
+
+data "pagerduty_event_orchestrations" "tf_my_monitor" {
+  search = ".*Orchestration$"
```
gsreynolds, line 24:
```suggestion
  name_filter = ".*Orchestration$"
```

website/docs/d/event_orchestrations.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `search` - (Required) The regex name of Global Event orchestrations to find in the PagerDuty API.
```
gsreynolds, line 33:
```suggestion
* `name_filter` - (Required) The regex name of Global Event orchestrations to find in the PagerDuty API.
```

website/docs/d/event_orchestrations.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+## Attributes Reference
+
+* `event_orchestrations` - The list of the Event Orchestrations which name match `search` argument.
```
gsreynolds, line 37:
```suggestion
* `name_filter` - The list of the Event Orchestrations which name match `name_filter` argument.
```

gsreynolds (changes requested):

gsreynolds (approved):
LGTM 🚀

imjaroiswebdev:
Awesome! Thank you @fgouteroux for adding this new data source and also @gsreynolds for all your help during the review, I really like the use case for this data source 💪🏽 🎉

gsreynolds:
Thank you @fgouteroux and @imjaroiswebdev!  🥳

fgouteroux:
Thank you so much for the review @gsreynolds and for your additional work @imjaroiswebdev !

imjaroiswebdev:
Hey @fgouteroux and @gsreynolds this new Data Source you introduced in now available on PagerDuty Terraform Provider [v2.10.0](https://github.com/PagerDuty/terraform-provider-pagerduty/releases/tag/v2.10.0) Thank you! 🎉 💪🏽 🚀

PR #605, opened 2022-12-23
replace ValidateValueFunc with ValidateValueDiagFunc due to deprecation

jedelson-pagerduty:
`schema.SchemaValidateFunc` has been deprecated and the recommendation is to use `schema.SchemaValidateDiagFunc` instead. This pull request replaces all usages of `schema.SchemaValidateFunc` with a diag-based function.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	cloud.google.com/go v0.71.0 // indirect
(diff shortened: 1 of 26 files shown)


imjaroiswebdev (approved):
This is great @jedelson-pagerduty thank you! This looks very good 💪🏽 💪🏽 🎉

PR #596, opened 2022-12-09
implementation of incident workflow and triggers

jedelson-pagerduty:
This adds support for the new PagerDuty Incident Workflows (both Data Source and Resource) and Incident Workflow Triggers (Resource only) features. These are currently in Early Access and not available to all accounts.

Acceptance tests are gated behind an environment variable `PAGERDUTY_ACC_INCIDENT_WORKFLOWS`. This can be set to any value, but `1` is recommended for consistency with `TF_ACC`:

```
$ PAGERDUTY_ACC_INCIDENT_WORKFLOWS=1 make testacc TESTARGS="-run PagerDutyIncidentWorkflow"                              
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run PagerDutyIncidentWorkflow -timeout 120m
?       github.com/terraform-providers/terraform-provider-pagerduty     [no test files]
=== RUN   TestAccDataSourcePagerDutyIncidentWorkflow
--- PASS: TestAccDataSourcePagerDutyIncidentWorkflow (7.46s)
=== RUN   TestAccDataSourcePagerDutyIncidentWorkflow_Missing
--- PASS: TestAccDataSourcePagerDutyIncidentWorkflow_Missing (1.46s)
=== RUN   TestAccPagerDutyIncidentWorkflow_import
--- PASS: TestAccPagerDutyIncidentWorkflow_import (7.93s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_import
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_import (14.32s)
=== RUN   TestAccPagerDutyIncidentWorkflow_Basic
--- PASS: TestAccPagerDutyIncidentWorkflow_Basic (10.26s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BadType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BadType (1.09s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithManualType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithManualType (1.21s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ConditionalTypeWithoutCondition
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ConditionalTypeWithoutCondition (1.39s)
[...]

PR #609, opened 2023-01-09
add team support for incident workflows

jedelson-pagerduty:
This is an addition to the support for incident workflows added in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/596. Workflows may optionally have a team assigned to them which limits which users can edit that workflow. See https://developer.pagerduty.com/api-reference/2d183500bb4fb-create-an-incident-workflow for more information.

As with #596 , acceptance test is guarded behind the `PAGERDUTY_ACC_INCIDENT_WORKFLOWS` environment variable.


imjaroiswebdev (approved):
Awesome @jedelson-pagerduty! 🚀 🎉

Issue #578, opened 2022-10-05
Error "Phone number should only contain digits"


JnMik:
Hello !
I am having issues applying my user contact method after running a terraform init --upgrade and it was working before.
So I have downgraded versions to identify that the bug seems to have appear in 2.4.0

Simply put, these 2 resources seems affected
![image](https://user-images.githubusercontent.com/4965516/194111123-cf2c8b75-9337-4950-85bd-3d2880089ab8.png)
![image](https://user-images.githubusercontent.com/4965516/194111181-9b89b614-cbba-432b-9d33-ee6ded24ea37.png)

### Terraform Version
Any version of the provider 2.4.0 seems affected

### Affected Resource(s)
pagerduty_user_contact_method using _phone_contact_method_ or _phone_contact_method_

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Debug Output
 Error: phone numbers should only contain digits

### Expected Behavior
User method should be applied properly when respecting the documentation format (country code with +1 and number as a digit wrapped in string double quotes). 

### Actual Behavior
Can't apply 

### References
I think the bug has been introduced here : https://github.com/PagerDuty/terraform-provider-pagerduty/pull/475/files

shgawas:
Phone number should contain maximum 10 digits. Just solved the small problem... !

Tzrlk:
I've got it as well on 2.9.1, with the following config:
```terraform
resource "pagerduty_user_contact_method" "mobile_text" {
  user_id      = local.user_id
  type         = "sms_contact_method"
  country_code = tonumber(local.mobile_prefix)
  address      = tonumber(local.mobile_number)
  label        = "Mobile"
}
```
and
```json
{
  "mobile_number": "208230116"
  "mobile_prefix": "64"
}
```
I've tried varying the length of the number to no effect.

Interestingly, if I put the values in as static inputs, or set the local vars to static inputs, it appears to work fine. It's exactly the same data, and the `tonumber` call should pick up if there were any issues with the content.

The code I'm using for the input parsing is:
```terraform
# Mobile number sanitising.
locals {

  # Clean up the input mobile by cleaning up any whitespace.
  mobile = replace(var.mobile, "\\s", "")

  # Extract the mobile prefix if it is possible.
  mobile_prefix = one(regex("^\\+?(${join("|", local.mobile_prefixes)})", local.mobile))

  # Strip the prefix from the mobile number if needed
[...]

PR #614, opened 2023-01-16
New test-cases to highlight possible phone inputs.

Tzrlk:
* Adding test-cases for existing validation errors.
* Reproducing #578.


Tzrlk:
As I don't have an active golang workspace, I'm relying on the automated build to check whether my blind changes are any good. One of the test cases is designed specifically to fail and will require guidance from the maintainers as to how it gets fixed.

PR #616, opened 2023-01-18
Fix go module name

### LINKED ISSUE #811: Incorrect Go Module Path
The go module path for this provider is listed as `github.com/terraform-providers/terraform-provider-pagerduty`:

https://github.com/PagerDuty/terraform-provider-pagerduty/blob/7133e1ff7561c406416bf2a971cdd77ffb014c79/go.mod#L1

The repository (and it's code) is hosted at `github.com/PagerDuty/terraform-provider-pagerduty`, a different org.

This makes it impossible to import any code in this provider using Go's normal mechanism.

Changing the import path here should not be a breaking change, since no one was previously able to import any code from this repo.

### Context

I work at @pulumi and one of my jobs is to maintain https://github.com/pulumi/pulumi-pagerduty. pulumi-pagerduty is a projection of `github.com/PagerDuty/terraform-provider-pagerduty`, so it includes parts of this repo as a library.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/PagerDuty/terraform-provider-pagerduty/pull/616 (fixes this issue)

iwahbe:
Go suggests that modules correspond with version control paths. Since terraform-provider-pagerduty is hosted under PagerDuty, go suggests calling the module github.com/PagerDuty/terraform-provider-pagerduty.

Fixes #811

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/terraform-providers/terraform-provider-pagerduty
+module github.com/PagerDuty/terraform-provider-pagerduty
 
 go 1.20
 
--- a/main.go
+++ b/main.go
@@ -8,8 +8,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server"
 	"github.com/hashicorp/terraform-plugin-mux/tf5muxserver"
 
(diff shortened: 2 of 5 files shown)


imjaroiswebdev (approved):
Thanks updating this @iwahbe 👍🏽

PR #598, opened 2022-12-14
[ORCA-3828] Return warnings when updating Event Orchestration Path

alenapan:
## Changes
This PR updates the following resources to display a list of warnings returned by the Event Orchestration Path Update endpoint to the Provider user:
- pagerduty_event_orchestration_router
- pagerduty_event_orchestration_unrouted
- pagerduty_event_orchestration_service

The following changes have been made to each of the aforementioned resources:
- Switched the resource to using context-aware methods (ReadContext, CreateContext, UpdateContext, DeleteContext). We need to use these methods in order to return a collection of [Diagnostics](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/diag)
- Updated signatures of the CRUD handler functions to accept `context.Context` as the first argument and return diag.Diagnostics (this is what the new methods require)
- Modified the bodies of the CRUD handler functions to return `diag.Diagnostics` instead of `error` or `nil`

## Before Merging
⚠️ We need to update the version of the API Client after https://github.com/heimweh/go-pagerduty/pull/105 is merged

### Acceptance Tests
- [x] Event Orchestration Acceptance Tests are passing
image

### Testing Event Orchestration Config with the Local Copy of the Provider

https://github.com/PagerDuty/terraform-provider-pagerduty-internal/pull/3

**Account that should see no warnings (used the personal account in prod)**

- [x] Creating a new service, orchestration, router, unrouted, and service paths works:
```cmd
Plan: 5 to add, 0 to change, 0 to destroy.
pagerduty_event_orchestration.orca_3828_test_tf_warning: Creating...
pagerduty_service.orca_3828: Creating...
pagerduty_service.orca_3828: Creation complete after 1s [id=P3RJ6G4]
pagerduty_event_orchestration.orca_3828_test_tf_warning: Creation complete after 1s [id=4399925b-3c90-44b5-a104-91b7b28b7475]
data.pagerduty_event_orchestration.data_orca_3828_test_tf_warning: Reading...
pagerduty_event_orchestration_router.orca_3828_test_tf_warning_router: Creating...
pagerduty_event_orchestration_unrouted.orca_3828_test_tf_warning_unrouted: Creating...
data.pagerduty_event_orchestration.data_orca_3828_test_tf_warning: Read complete after 0s [id=4399925b-3c90-44b5-a104-91b7b28b7475]
[...]

PR #618, opened 2023-01-23
Event Orchestration Updates: Orchestration Warnings, Global Orchestrations, Orchestration Integrations

alenapan:
### Orchestration Warnings (ORCA-3828)
Updates the following resources to display a list of warnings returned by the Event Orchestration Path Update endpoint to the Provider user:
- pagerduty_event_orchestration_router
- pagerduty_event_orchestration_unrouted
- pagerduty_event_orchestration_service

The following changes have been made to each of the aforementioned resources:
- Switched the resource to using context-aware methods (ReadContext, CreateContext, UpdateContext, DeleteContext). We need to use these methods in order to return a collection of [Diagnostics](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/diag)
- Updated signatures of the CRUD handler functions to accept `context.Context` as the first argument and return diag.Diagnostics (this is what the new methods require)
- Modified the bodies of the CRUD handler functions to return `diag.Diagnostics` instead of `error` or `nil`

**TESTING**
See https://github.com/PagerDuty/terraform-provider-pagerduty/pull/598

### Global Orchestrations (ORCA-3893 & ORCA-3895)
Adds the new `pagerduty_event_orchestration_global` resource, acceptance tests, and documentation

**TESTING**
See https://github.com/PagerDuty/terraform-provider-pagerduty/pull/622

### Orchestration Integrations (ORCA-3946 & ORCA-3947)
Adds the new `pagerduty_event_orchestration_integration` resource, acceptance tests, and documentation

**TESTING**
See https://github.com/PagerDuty/terraform-provider-pagerduty/pull/640

## ⚠️ Before Merging

- [ ] update the version of the API Client after https://github.com/heimweh/go-pagerduty/pull/118 is merged
- [x] update URLs in the Orchestration Integration datasource doc after the Public API documentation is released


pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+				if readDestErr != nil {
+					log.Printf("[WARN] Integration '%s' cannot be found on the destination PagerDuty Event Orchestration '%s'. Retrying migration....", id, destinationOrchId)
+					return resource.RetryableError(readDestErr)
+				}
+
+				// retry migration if the integration still exists on the source:
+				if readSrcErr == nil && srcInt != nil {
+					log.Printf("[WARN] Integration '%s' still exists on the source PagerDuty Event Orchestration '%s'. Retrying migration....", id, sourceOrchId)
+					return resource.RetryableError(fmt.Errorf("Integration '%s' still exists on the source PagerDuty Event Orchestration '%s'.", id, sourceOrchId))
+				}
```
alenapan, line 186: TODO: refactor to a separate functions with retries (see pagerduty/resource_pagerduty_team_membership.go)

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+			log.Printf("[INFO] Updating Integration '%s' for PagerDuty Event Orchestration: %s", id, oid)
+
+			if integration, _, err := client.EventOrchestrationIntegrations.UpdateContext(ctx, oid, id, payload); err != nil {
+				if isErrCode(err, 400) {
+					return resource.NonRetryableError(err)
+				}
+				return resource.RetryableError(err)
+			} else if integration != nil {
+				// Try reading an integration after updating the label, retry if the label is not updated:
+				if updInt, _, readErr := client.EventOrchestrationIntegrations.GetContext(ctx, oid, id); readErr == nil && updInt.Label != payload.Label {
```
alenapan, line 211: TODO: replace with Fetch.. function

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+
+	id := d.Id()
+	oid, _ := getEventOrchestrationIntegrationPayloadData(d)
+
+	retryErr := resource.RetryContext(ctx, 2*time.Minute, func() *resource.RetryError {
+		log.Printf("[INFO] Deleting Integration '%s' for PagerDuty Event Orchestration: %s", id, oid)
+		if _, err := client.EventOrchestrationIntegrations.DeleteContext(ctx, oid, id); err != nil {
+			return resource.RetryableError(err)
+		} else {
+			// Try reading an integration after deletion, retry if still found:
```
alenapan, line 244: TODO: replace with Fetch function

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+		return diag.FromErr(retryErr)
+	}
+
+	d.SetId("")
+
+	return nil
+}
+
+func resourcePagerDutyEventOrchestrationIntegrationImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	ids := strings.Split(d.Id(), ".")
```
alenapan, line 264: Use colon separated IDs and update docs and tests, use `resourcePagerDutyParseColonCompoundID` function

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+		log.Printf("[INFO] Reading Integration '%s' for PagerDuty Event Orchestration: %s", id, oid)
+
+		if integration, _, err := client.EventOrchestrationIntegrations.GetContext(ctx, oid, id); err != nil {
+			return resource.RetryableError(err)
+		} else if integration != nil {
+			d.SetId(id)
+			d.Set("event_orchestration", oid)
+			setEventOrchestrationIntegrationProps(d, integration)
+		}
+		return nil
```
alenapan, line 285: use fetch function

pagerduty/resource_pagerduty_event_orchestration_path_router.go
```diff
@@ -110,13 +112,15 @@ func resourcePagerDutyEventOrchestrationPathRouter() *schema.Resource {
 	}
 }
 
-func resourcePagerDutyEventOrchestrationPathRouterRead(d *schema.ResourceData, meta interface{}) error {
+func resourcePagerDutyEventOrchestrationPathRouterRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
alenapan, line 115: Add contexrt methods to the API client, use RetryContext

pagerduty/resource_pagerduty_event_orchestration_path_service.go
```diff
@@ -213,27 +176,30 @@ func resourcePagerDutyEventOrchestrationPathService() *schema.Resource {
...
 			time.Sleep(2 * time.Second)
 			return resource.RetryableError(err)
 		}
+
 		return nil
 	})
+
 	if retryErr != nil {
-		return retryErr
+		return diag.FromErr(retryErr)
```
alenapan, line 202: Also add context-aware methods for status get/update endpoint

pagerduty/import_pagerduty_event_orchestration_path_global_test.go
```diff
@@ -0,0 +1,38 @@
+package pagerduty
```
alenapan, line 1: TODO: make sure tests aren't failing for accounts without proper entitlements

pagerduty/resource_pagerduty_event_orchestration_integration.go
alenapan: Addressed in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/618/commits/9c7ddd0faf4d463f7dd1118d14936863a7232114

pagerduty/resource_pagerduty_event_orchestration_path_router.go
alenapan: Addressed in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/618/commits/9c7ddd0faf4d463f7dd1118d14936863a7232114

pagerduty/resource_pagerduty_event_orchestration_path_service.go
alenapan: Addressed in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/618/commits/9c7ddd0faf4d463f7dd1118d14936863a7232114

pagerduty/import_pagerduty_event_orchestration_path_global_test.go
alenapan: Unfortunately it's hard to test warnings because in order to make features ineligible we need to use tokens for an account that doesn't have these features.

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,311 @@
...
+		} else {
+			d.SetId(id)
+			d.Set("event_orchestration", oid)
+			setEventOrchestrationIntegrationProps(d, integration)
+			return integration, nil
+		}
+	}
+
+	return nil, fmt.Errorf("Reading Integration '%s' for PagerDuty Event Orchestration '%s' returned `nil`.", id, oid)
+}
```
alenapan, line 105: This function doesn't have retries in it because it's called from inside `.RetryContext()` methods that implement retries. Also specifically in case with integrations if we can't read the created/updated integration the first time it means that the integration wasn't created/updated so reading it repeatedly won't make any difference. Instead we want to fail fast and retry creating/updating an integration.

pagerduty/resource_pagerduty_event_orchestration_integration.go
alenapan: I used `.GetContext()` here directly instead of calling `fetchPagerDutyEventOrchestrationIntegration()` because when we perform a migration we don't need to set any integration properties that the fetch function does, especially when we are verifying that the integration doesn't exist on the source.

pagerduty/resource_pagerduty_event_orchestration_integration.go
alenapan: Same here, I left the call to `.GetContext()` instead of using the fetch function because we don't need to set any of the integration properties when we delete an integration resource, we are just retrying the deletion process if the integration still exists.

alenapan:
Tested manually and ran acceptance tests after the latest commit:

**Acceptance tests**
image

**Manual Testing**

- [x] Create/read/update delete Event Orchestration (`pagerduty_event_orchestration`)
- [x] Create/read/update delete Global Orchestration (`pagerduty_event_orchestration_global`)
- [x] Create/read/update delete Router Orchestration  (`pagerduty_event_orchestration_router`)
- [x] Create/read/update delete Unrouted Orchestration  (`pagerduty_event_orchestration_unrouted`)
- [x] Create/read/update delete Service Orchestration  (`pagerduty_event_orchestration_service`)
- [x] Create/read/update delete Event Orchestration Integration (`pagerduty_event_orchestration_integration`)
- [x] Import Event Orchestration Integration using the `:` separator

PR #621, opened 2023-01-25
EF-3964 Address `team_membership` inconsistency after `create` and `update`

imjaroiswebdev:
Addresses the issue with `pagerduty_team_membership` that causes a permanent diff on TF state with the user **role**. So now the introduction of a few conditional retries to the GET call to validate the change when needed, the state diff can be remediated.

## Test cases introduced...
Screenshot 2023-01-25 at 17 26 42


johncoleman83 (approved):
Changes look great!

PR #622, opened 2023-01-27
[ORCA-3893] Add support for Global Orchestrations

alenapan:
## Changes

⚠️ Depends on the changes to the Go API client: https://github.com/alenapan/go-pagerduty/pull/11

- Add a new `pagerduty_event_orchestration_global` resource
- Extract Automation Actions (Webhooks) schemas and expand / flatten method into shared functions used by`pagerduty_event_orchestration_global` and `pagerduty_event_orchestration_service` resources
- Add acceptance tests for the new resource

### Acceptance Tests

Event Orchestration Acceptance Tests are passing:

image

### Testing Event Orchestration Config with the Local Copy of the Provider

Event Orchestration: https://pdt-alenatestaccount.pagerduty.com/event-orchestration/f8194217-5624-4f69-bdb7-788d52c8277f/global-rules

| Terraform Config | S3 Object Version | Works? |
| --- | --- | --- |
| [Empty State](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L55) | tCceWlWRP.f2eIFLhMAwpbwRhnF6vj8j | ✅ |
| [Full Config](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L67) | P08rcGy3aEiN9TmepieiCeTgyazDjRbL | ✅ |
| [Full Config - Update PD Automation Actions Params](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L197) | __3xtHA_FD2XOqQCahXK1kCZe0SU64hb | ✅  |
| [Full Config - Delete PD Automation Actions Params](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L326) | 5W3kIEwiqyGxjR6x3kPg4YiClQE1tUaR | ✅  |
| [Full Config Updated](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L437) | k.EraGwccwYqd1TQkz5_kDHEt_1aVf0e | ✅  |
| [Full Config - Rule Actions/Conditions Deleted](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L559) | YFKkbF92qGCR3L9IRMbnYw_URNvFI567 | ✅  |
| [Full Config - Rules Deleted](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L591) | oV02KuPTn_XJKMPAfhiNTo.BiWsZshmK | ✅  |
| [Full Config - Sets Deleted](https://github.com/PagerDuty/terraform-provider-pagerduty-internal/blob/ORCA-3893-global-orch-test/main.tf#L615) | 7CDlM6x.27QJNkZXdnIC9tNxBWxSRVSh | ✅  |

### DIFF
--- a/pagerduty/event_orchestration_path_util.go
+++ b/pagerduty/event_orchestration_path_util.go
@@ -54,6 +54,47 @@ var eventOrchestrationPathExtractionsSchema = map[string]*schema.Schema{
 	},
 }
 
+var eventOrchestrationAutomationActionObjectSchema = map[string]*schema.Schema{
+	"key": {
+		Type:     schema.TypeString,
+		Required: true,
+	},
+	"value": {
+		Type:     schema.TypeString,
+		Required: true,
+	},
+}
+
+var eventOrchestrationAutomationActionSchema = map[string]*schema.Schema{
+	"name": {
+		Type:     schema.TypeString,
+		Required: true,
+	},
+	"url": {
+		Type:     schema.TypeString,
+		Required: true,
+	},
+	"auto_send": {
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+	},
+	"header": {
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: eventOrchestrationAutomationActionObjectSchema,
+		},
+	},
+	"parameter": {
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: eventOrchestrationAutomationActionObjectSchema,
+		},
+	},
+}
+
 func invalidExtractionRegexTemplateNilConfig() string {
 	return `
 		extraction {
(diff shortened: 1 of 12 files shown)


pagerduty/event_orchestration_path_util.go
```diff
@@ -54,6 +54,47 @@ var eventOrchestrationPathExtractionsSchema = map[string]*schema.Schema{
 	},
 }
 
+var eventOrchestrationAutomationActionObjectSchema = map[string]*schema.Schema{
```
alenapan, line 57: The schemas and methods below are shared between Service and Global orchestrations so I moved them here from`pagerduty/resource_pagerduty_event_orchestration_path_service.go`

pagerduty/resource_pagerduty_event_orchestration_path_global.go
```diff
@@ -0,0 +1,423 @@
...
+	a := eventOrchestrationPathGlobalCatchAllActionsSchema
+	a["route_to"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Optional: true,
+	}
+
+	return a
+}
+
+func resourcePagerDutyEventOrchestrationPathGlobal() *schema.Resource {
```
alenapan, line 81: Although the schema for Global Orchestrations mostly overlaps with the Service Orchestration resource it's still slightly different (has `drop_event` but has no `pagerduty_automation_actions`). I thought that having both schemas inheriting from the same base schema would not make the code any more legible so i kept them separate. What I am sharing here is the Automation Action (webhook) schema that's more complex.

pagerduty/resource_pagerduty_event_orchestration_path_global_test.go
```diff
@@ -0,0 +1,739 @@
+package pagerduty
```
alenapan, line 1: The test steps are very similar to those in `pagerduty/resource_pagerduty_event_orchestration_path_service_test.go`, we make sure we can add/update/delete individual properties.

pagerduty/resource_pagerduty_event_orchestration_path_global.go
```diff
@@ -0,0 +1,423 @@
...
+
+	client, err := meta.(*Config).Client()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	retryErr := resource.Retry(2*time.Minute, func() *resource.RetryError {
+		id := d.Id()
+		t := "global"
+		log.Printf("[INFO] Reading PagerDuty Event Orchestration Path of type %s for orchestration: %s", t, id)
```
metavida, line 175: **Not a blocker:** Comparing this file with the `*_path_service.go` equivalent made me realize we've got a little typo in the "Reading PagerDuty Event Orchestration Path" log line for a Service Orchestration! Currently the log would read "of type service for orchestration: POOPBUG" ([here](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/a6e7827bd4c268dd76c687a8e2e1918fd7c52c3b/pagerduty/resource_pagerduty_event_orchestration_path_service.go#L184)) but it should *probably* read "of type service for service: POOPBUG" since the ID is the the Service that the Orchestration Path belongs to.

metavida (approved):
I like the tradeoffs you made between legibility vs repetition. Thanks for calling out that decision explicitly! A single non-blocking comment about specificity in a log line. Seems solid!

pagerduty/resource_pagerduty_event_orchestration_path_global.go
alenapan: Good catch, updated the message in [4f12ba4](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/622/commits/4f12ba4ef2e054b95c5d738ee2506423c1b54d6a)

PR #624, opened 2023-01-31
service: Compute additional fields already included in API response

lexton:
## WHAT

* Adds additional fields already present in the data model

## WHY

* We are trying to generate terraform + matching state import scripts for existing resources, this will allow us to model all the required input fields for the `pagerduty_service` resource to allow us to import existing into state successfully

## Tests

* All existing unit tests pass, and it is purely additive, if you want extra tests for each field - just let me know.
* I didn't run the acceptance tests locally as I don't have an environment setup to run them in - LMK if this is required.

### DIFF
--- a/pagerduty/data_source_pagerduty_service.go
+++ b/pagerduty/data_source_pagerduty_service.go
@@ -19,6 +19,43 @@ func dataSourcePagerDutyService() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"auto_resolve_timeout": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"acknowledgement_timeout": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"alert_creation": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"escalation_policy": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"teams": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "The set of teams associated with the service",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
 			"type": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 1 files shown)


MrColeC:
Thanks for raising this contribution @lexton 

Our goal with this contribution is to make it possible, or more possible, to import existing PagerDuty configurations into Terraform when they originated not from Terraform. With these fields missing in the data model, we are having real struggles on putting configurations under PagerDuty TF control.

If I missed anything, please correct me @lexton

pagerduty/data_source_pagerduty_service.go
```diff
@@ -19,6 +19,49 @@ func dataSourcePagerDutyService() *schema.Resource {
...
+				// The workaround described in SDK documentation is to confirm the required keys are set when expanding the Map object inside the resource code.
+				// See https://www.terraform.io/docs/extend/schemas/schema-types.html#typemap
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"teams": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "List of users who are members of the team",
```
jedelson-pagerduty, line 51: This description doesn't look right -- matching the API docs it should be `The set of teams associated with the service`

lexton: Fixed

pagerduty/data_source_pagerduty_service.go
```diff
@@ -19,6 +19,49 @@ func dataSourcePagerDutyService() *schema.Resource {
...
+			},
+			"alert_creation": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"escalation_policy": {
```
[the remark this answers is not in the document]
lexton: The upstream response body contains a lot more metadata than just the ID. I have collapsed that response down to just the ID, just above the setter.

pagerduty/data_source_pagerduty_service.go
jedelson-pagerduty: Thanks

pagerduty/data_source_pagerduty_service.go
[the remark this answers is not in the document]
jedelson-pagerduty: Makes sense now. Thanks.

lexton:
Any other changes we want to this?

imjaroiswebdev:
Hi @lexton thank you for this contribution, because I can understand the value in your use case for supporting this additional fields to the Service Data Source, however I would ask you to add an additional acceptance test that validates the correct assignment of each one of this new fields, particularly the `teams` one. On top of that, I would also ask you to please update the Service Data Source documentation to reflect these new fields on the **Attributes Reference** section of it. I believe that will be enough to merge your contribution. Please let me know if you need help to run or add the acceptance tests, I understand you don't have an environment setup for executing them.

PR #640, opened 2023-02-24
[ORCA-3941] Add support for Event Orchestration Integrations

alenapan:


pagerduty/provider.go
```diff
@@ -46,27 +46,28 @@ func Provider() *schema.Provider {
...
+			"pagerduty_extension_schema":                dataSourcePagerDutyExtensionSchema(),
+			"pagerduty_service":                         dataSourcePagerDutyService(),
+			"pagerduty_service_integration":             dataSourcePagerDutyServiceIntegration(),
+			"pagerduty_business_service":                dataSourcePagerDutyBusinessService(),
+			"pagerduty_priority":                        dataSourcePagerDutyPriority(),
+			"pagerduty_ruleset":                         dataSourcePagerDutyRuleset(),
+			"pagerduty_tag":                             dataSourcePagerDutyTag(),
+			"pagerduty_event_orchestration":             dataSourcePagerDutyEventOrchestration(),
+			"pagerduty_event_orchestrations":            dataSourcePagerDutyEventOrchestrations(),
+			"pagerduty_event_orchestration_integration": dataSourcePagerDutyEventOrchestrationIntegration(),
```
alenapan, line 65: Added `pagerduty_event_orchestration_integration` to the list and reformatted the block.

pagerduty/resource_pagerduty_event_orchestration.go
```diff
@@ -207,22 +211,14 @@ func flattenEventOrchestrationIntegrations(eoi []*pagerduty.EventOrchestrationIn
...
 
-func flattenEventOrchestrationIntegrationParameters(p *pagerduty.EventOrchestrationIntegrationParameters) []interface{} {
-	result := map[string]interface{}{
-		"routing_key": p.RoutingKey,
-		"type":        p.Type,
-	}
-
-	return []interface{}{result}
-}
-
```
alenapan, line 225: Moved this to pagerduty/resource_pagerduty_event_orchestration_integration.go

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+
+		diags = append(diags, diag.Diagnostic{
+			Severity:      diag.Warning,
+			Summary:       "Modifying the event_orchestration property of the 'pagerduty_event_orchestration_integration' resource will cause all future events sent with this integration's routing key to be evaluated against the new Event Orchestration.",
+			AttributePath: p,
+		})
+
+		return diags
+	}
+}
```
alenapan, line 72: The warning is always shown on `terraform plan`

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+			return resource.RetryableError(err)
+		} else if integration != nil {
+			// Try reading an integration after creation, retry if not found:
+			if _, _, readErr := client.EventOrchestrationIntegrations.GetContext(ctx, oid, integration.ID); readErr != nil {
+				log.Printf("[WARN] Cannot locate Integration '%s' on PagerDuty Event Orchestration '%s'. Retrying creation...", integration.ID, oid)
+				return resource.RetryableError(readErr)
+			} else {
+				d.SetId(integration.ID)
+				setEventOrchestrationIntegrationProps(d, integration)
+			}
```
alenapan, line 108: Concurrent requests to the same orchestration cause a race condition, that's why here and below we make another request to confirm that an integration was indeed created/updated/deleted, and retry if it wasn't. We have similar logic for event rules.

pagerduty/resource_pagerduty_event_orchestration_integration.go
```diff
@@ -0,0 +1,309 @@
...
+	})
+
+	if retryErr != nil {
+		return diag.FromErr(retryErr)
+	}
+
+	return nil
+}
+
+func resourcePagerDutyEventOrchestrationIntegrationUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
alenapan, line 149: If the `event_orchestration` has changed we make a call to the `/migration` endpoint to migrate the integration from one EO to another.

If the label attribute has changed we make an Update Integration call to update the label either in the source or the destination orchestration.

c-kaieong (approved):
Nothing looks dangerous or disastrous, from the perspective of someone that has zero experience in golang and writing TF providers. Domain-wise, LGTM

Hopefully a more thorough review with José will help catch things I'm not able to!

PR #641, opened 2023-02-24
Fix/slack url eu

MaxRBosch:
**What:**
Method to override Slack URL for EU users of pagerduty. If left empty same URL for Slack requests is used as before. This is therefore purely a way for users to override it and shouldnt impact others. 
**Why:**
Noticed that for our EU PagerDuty instance the slack_connection failed. This is because the URL to send requests to is not [app.pagerduty](https://app.pagerduty.com/) but should be `https://(pagerduty_instance).eu.pagerduty.com` as you can see that requests are made to this address when creating Slack connections in the PagerDuty UI.
**Tests:**
Tests are passing

### DIFF
--- a/pagerduty/config.go
+++ b/pagerduty/config.go
@@ -29,6 +29,9 @@ type Config struct {
 	// The PagerDuty User level token for Slack
 	UserToken string
 
+	// The URL to send requests to Slack
+	SlackUrlOverride string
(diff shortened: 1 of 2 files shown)


MaxRBosch:
This is fixed so this feature is not needed

Issue #565, opened 2022-09-01
Missing support for Service Orchestration active status for a Service


etiennechabert:
## Terraform Version
2.6.1

## Affected Resource(s)
[service](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service)

## Missing feature 

There is no way to change for a given service the Service Orchestration Status, while it's possible to automate everything else, like:
- [service](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service)
- [event_orchestration_service](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/event_orchestration_service)

Problem: We have hundreds of services for which we have via terraform automated the definition of orchestration rules, and it seems that we have now to enable the service orchestration for each of them manually? 

Does not make much sense, and a bit frustrating since this API endpoint seems to be doing the trick: https://developer.pagerduty.com/api-reference/855659be83d9e-update-the-service-orchestration-active-status-for-a-service

It would be really beneficial if this could be included into the `service` pagerduty module, so the user can define if he wants to use `service ruleset` (should probably be the default) or `service orchestration path`

Thanks in advance

etiennechabert:
For the one that might be blocked by this... here is my workaround

```terraform
resource "pagerduty_service" "default" {...}

resource "null_resource" "service_event_orchestration_activer" {
  provisioner "local-exec" {
    command = "curl --request PUT --url https://api.pagerduty.com/event_orchestrations/services/${pagerduty_service.default.id}/active --header 'Accept: application/vnd.pagerduty+json;version=2' --header 'Authorization: Token token=${var.pagerduty_token}' --header 'Content-Type: application/json' --data '{\"active\": true}'"
  }

  depends_on = [pagerduty_service.default]
}
```

marco-hoyer:
Hey Pagerduty folks. This actually is a big issue for us. Any idea when this could be fixed?

marco-hoyer:
Hey @NargiT. Just FYI Pagerduty seemed to have a bigger issue going on affecting many customers. This was fixed somewhen last week but we still had the issue. A fix for it was deployed just yesterday after we had to spam the support quite a bit. So either the problem is solved for you as well or I'd suggest you also ask the support about that a couple times.

etiennechabert:
@imjaroiswebdev I am really sorry for tagging you directly, doing so because:
[...]

Issue #643, opened 2023-02-27
Activate Event Orchestration for a service via terraform


AstritCepele:
Hi there,

### Terraform Version
❯ terraform -v
Terraform v1.3.6
+ provider registry.terraform.io/pagerduty/pagerduty v2.11.0

### Affected Resource(s)
Please list the resources as a list, for example:
- pagerduty_service
- pagerduty_event_orchestration_service

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "pagerduty_team" "engineering" {
  name = "Engineering"
}

resource "pagerduty_user" "example" {
  name  = "Earline Greenholt"
  email = "125.greenholt.earline@graham.name"
  teams = [pagerduty_team.engineering.id]
}

resource "pagerduty_escalation_policy" "foo" {
  name      = "Engineering Escalation Policy"
  num_loops = 2

  rule {
    escalation_delay_in_minutes = 10
    target {
      type = "user"
      id   = pagerduty_user.example.id
    }
  }
}

resource "pagerduty_service" "example" {
  name                    = "My Web App"
  auto_resolve_timeout    = 14400
  acknowledgement_timeout = 600
  escalation_policy       = pagerduty_escalation_policy.example.id
  alert_creation          = "create_alerts_and_incidents"
}

data "pagerduty_priority" "p1" {
  name = "P1"
}

resource "pagerduty_event_orchestration_service" "www" {
  service = pagerduty_service.example.id
  set {
    id = "start"
    rule {
      label = "Always apply some consistent event transformations to all events"
      actions {
        variable {
          name = "hostname"
          path = "event.component"
          value = "hostname: (.*)"
          type = "regex"
        }
        extraction {
          # Demonstrating a template-style extraction
          template = "{{variables.hostname}}"
          target = "event.custom_details.hostname"
        }
        extraction {
          # Demonstrating a regex-style extraction
          source = "event.source"
          regex = "www (.*) service"
          target = "event.source"
        }
        # Id of the next set
        route_to = "step-two"
      }
    }
  }
  set {
    id = "step-two"
    rule {
      label = "All critical alerts should be treated as P1 incident"
      condition {
        expression = "event.severity matches 'critical'"
      }
      actions {
        annotate = "Please use our P1 runbook: https://docs.test/p1-runbook"
        priority = data.pagerduty_priority.p1.id
      }
    }
    rule {
[...]

Issue #589, opened 2022-11-11
Feature Request: Switch to Service Orchestrations


mbelang:
### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.
1.2.9

### Affected Resource(s)
- pagerduty_service

### Terraform Configuration Files
```hcl
resource "pagerduty_service" "this" {
  name              = var.configuration.name
  description       = var.configuration.description
  alert_creation    = "create_alerts_and_incidents"
  escalation_policy = var.configuration.default_escalation_policy_id

  incident_urgency_rule {
    type = "constant"

    urgency = var.configuration.urgency
  }

  auto_resolve_timeout    = "null"
  acknowledgement_timeout = "null"
}
```

### Expected Behavior
Being able to `Switch to Service Orchestrations` from terraform
![Screenshot from 2022-11-11 13-43-32](https://user-images.githubusercontent.com/5888932/201410239-cd907485-51dd-4108-98d1-84f93b87901a.png)

Add an optional `event_management` argument with default to `basic` and optional `orchestration`

example: 

```hcl
resource "pagerduty_service" "this" {
  name              = var.configuration.name
  description       = var.configuration.description
  alert_creation    = "create_alerts_and_incidents"
  escalation_policy = var.configuration.default_escalation_policy_id
  
  event_management = "orchestration"

  incident_urgency_rule {
    type = "constant"

    urgency = var.configuration.urgency
  }

  auto_resolve_timeout    = "null"
  acknowledgement_timeout = "null"
}
```

### Actual Behavior
Not possible to switch the service to orchestration via terraform option in the service resource (or it is not documented). It is only possible to switch from the UI.

etiennechabert:
Maybe a duplicate with: https://github.com/PagerDuty/terraform-provider-pagerduty/issues/565

+ Here you have a poor workaround, maybe will help: https://github.com/PagerDuty/terraform-provider-pagerduty/issues/565#issuecomment-1239345555

metavida:
Confirmed that you now **can** switch any service to use Event Orchestration using [the `enable_event_orchestration_for_service.enable_event_orchestration_for_service` argument](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/event_orchestration_service#enable_event_orchestration_for_service) as of version 2.11.1 of the PagerDuty Terraform Provider.

PR #649, opened 2023-03-08
Enable event orchestration active status for service

imjaroiswebdev:
Closes: #565 #589 #643 

This update will bring support for the new attribute `enable_event_orchestration_for_service` to the resource `pagerduty_event_orchestration_service` with the direct effect of allowing to switch to Service Orchestration using this new attribute from the EO Service Terraform configuration instead of having to make it through the PagerDuty's UI.

## Acceptance Criteria

- [x] `enable_event_orchestration_for_service` is a property of the pagerduty_event_orchestration_service resource.
- [x] The property is boolean, optional, no default value.
- [x] We will need to always call this endpoint in the provider’s Read handler function to set the property.
- [x] If the user doesn’t specify the property on the resource we do not need to call this endpoint in the Update handler function.
- [x] Handle **410 (Gone)** response from EO Service Active status endpoint.

## Example of use...

```hcl
resource "pagerduty_event_orchestration_service" "serviceA" {
  service                                = pagerduty_service.example.id
  enable_event_orchestration_for_service = true

  set {
    id = "start"
  }

  catch_all {
    actions {}
  }
}
```

Depends on: https://github.com/heimweh/go-pagerduty/pull/127

### DIFF
--- a/pagerduty/resource_pagerduty_event_orchestration_path_service.go
+++ b/pagerduty/resource_pagerduty_event_orchestration_path_service.go
@@ -1,7 +1,9 @@
 package pagerduty
 
 import (
+	"fmt"
 	"log"
+	"net/http"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
@@ -139,6 +141,10 @@ func resourcePagerDutyEventOrchestrationPathService() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"enable_event_orchestration_for_service": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
 			"set": {
 				Type:     schema.TypeList,
 				Required: true,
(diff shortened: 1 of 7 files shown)


pagerduty/resource_pagerduty_event_orchestration_path_service.go
```diff
@@ -240,29 +277,69 @@ func resourcePagerDutyEventOrchestrationPathServiceUpdate(d *schema.ResourceData
...
+func needToUpdateServiceActiveStatus(d *schema.ResourceData) bool {
+	var needToUpdate bool
+	if d.HasChange("enable_event_orchestration_for_service") {
+		o, n := d.GetChange("enable_event_orchestration_for_service")
+		old := o.(bool)
+		new := n.(bool)
+		_, ok := d.GetOkExists("enable_event_orchestration_for_service")
+		if ok || old != new && new == false {
+			needToUpdate = true
+		}
```
alenapan, line 337: Question (just to make sure I understand what this code does) - does the condition `ok || old != new && new == false` mean that `needToUpdate` will be set to `true` in the following cases:
- if a user has `enable_event_orchestration_for_service` explicitly set to either `true` or `false`
- if a user had `enable_event_orchestration_for_service = true` and then deleted it from their TF config?

In other words if the config looks like this:
```hcl
resource "pagerduty_event_orchestration_service" "serviceA" {
    service = "POOPBUG"
    enable_event_orchestration_for_service = true
    ...
}
```
**OR**:
```hcl
resource "pagerduty_event_orchestration_service" "serviceA" {
    service = "POOPBUG"
    enable_event_orchestration_for_service = false
    ...
}
```

**OR**:
```hcl
# old config:
resource "pagerduty_event_orchestration_service" "serviceA" {
    service = "POOPBUG"
    enable_event_orchestration_for_service = true
    ...
}

# new config:
resource "pagerduty_event_orchestration_service" "serviceA" {
    service = "POOPBUG"
    #  enable_event_orchestration_for_service was deleted
    ...
}
```

alenapan (commented):
Left one question to confirm the logic but otherwise LGTM 👍

pagerduty/resource_pagerduty_event_orchestration_path_service.go
alenapan: Sounds good, thanks for confirming!

alenapan (approved):
LGTM 👍 Thank you very much for adding this feature!

Issue #389, opened 2021-09-14
resource pagerduty_service ignoring lifecycle rule for last_incident_timestamp


davekaufman:
### Terraform Version
- `Terraform v1.0.0`

pagerduty provider version: 
- `Installed pagerduty/pagerduty v1.9.9 (signed by a HashiCorp partner, key ID 027C6DD1F0707B45)`

### Affected Resource(s)
``` 
# module.external_services_plaid.pagerduty_service.this has been changed
~ resource "pagerduty_service" "this" {
        id                      = "PG2RQLF"
      ~ last_incident_timestamp = "2021-09-05T16:16:07-07:00" -> "2021-09-09T17:30:16-07:00"
        name                    = "External services Plaid (Service)"
        # (8 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

### Terraform Configuration Files
```hcl
resource "pagerduty_service" "this" {
  name                    = "${var.service_name} (Service)"
  escalation_policy       = var.escalation_policy.id
  alert_creation          = "create_alerts_and_incidents"
  acknowledgement_timeout = "null"
  auto_resolve_timeout    = "null"

  lifecycle {
    ignore_changes = [last_incident_timestamp, status]
  }
}
```

### Expected Behavior
per the `lifecycle` block on the resource, changes to `last_incident_timestamp` should have been ignored.

### Actual Behavior
changes to `last_incident_timestamp` result in copious irrelevant plan output

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
pretty bog-standard

### References
search for `last_incident_timestamp` in previous issues returned no results.

bradyclifford:
Seeing this well.

davidferdinand:
Seeing this too, very annoying with the unwanted plan changes

idsvandermolen:
Similarly we also see a outside of Terraform change of `status` in the `pagerduty_service` rule.
```
 # module.platform-service.pagerduty_service.service has been changed
  ~ resource "pagerduty_service" "service" {
        id                      = "PAQCMT4"
      + last_incident_timestamp = "2021-12-02T10:18:57+01:00"
        name                    = "Platform"
      ~ status                  = "active" -> "critical"
        # (8 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

```

awilson1801:
Any update on this?

cep21:
[...]

PR #690, opened 2023-05-24
feat: move last_incident_timestamp and status to data source only

SamKirsch10:
Fixes issue #389. These two attributes will change over time, making terraform show constant `Terraform detected the following changes made outside of Terraform since the last "terraform apply" which may have affected this plan`. 

This proposal is to move `last_incident_timestamp` and `status` out of the Service resource and into the data Service source. This will allow anyone to query these attributes while maintaining the information of the state = reality concept for terraform.

### DIFF
--- a/pagerduty/data_source_pagerduty_service.go
+++ b/pagerduty/data_source_pagerduty_service.go
@@ -24,11 +24,19 @@ func dataSourcePagerDutyService() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"last_incident_timestamp": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"acknowledgement_timeout": {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
 			"alert_creation": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"status": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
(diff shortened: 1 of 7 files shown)


SamKirsch10:
@jedelson-pagerduty - That totally makes sense. Let me add that in.

SamKirsch10:
@imjaroiswebdev and @jedelson-pagerduty  what do you guys think?

SamKirsch10:
bump @imjaroiswebdev and @jedelson-pagerduty what do you guys think?

rodrigopinto:
Hey folks, thanks for the work on this! Any update on what is needed to move it to a merged state?

bendrucker:
Left some notes on https://github.com/PagerDuty/terraform-provider-pagerduty/issues/389#issuecomment-1666129305. The key points:

* Computed attributes that change on the server are common and this "state = reality" concept a few folks have referenced is not so cut and dry.
* Either this attribute should exist on both the resource and data source, or neither. Neither and a dedicated `pagerduty_service_status` data source would fit with Terraform conventions too.
* Filling these fields with garbage data is a breaking change, as breaking as removing them. Arguably more so, because at least the latter is explicit and guarantees an error.

Happy to explain further on the issue. Just seemed like some threads were going off track here and I was here anyway to fix a conflict in another PR.

SamKirsch10:
@bendrucker - That totally makes sense. I believe a separate data source then is the solution here. I've made the changes and caught up this branch to latest release

PR #701, opened 2023-06-15
Custom Fields - remove early access marker from incident custom fields pages

jedelson-pagerduty:
Incorporates https://github.com/heimweh/go-pagerduty/pull/134

as well as updating the documentation pages now that Custom Fields is Generally Available

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.15.0
 	github.com/hashicorp/terraform-json v0.13.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.10.1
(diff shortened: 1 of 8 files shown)


imjaroiswebdev (changes requested):
@jedelson-pagerduty thank you for the quick response to remove the markers on docs. However, before moving forward with the merge,  https://github.com/heimweh/go-pagerduty/pull/134 was merged and therefore You would need to update the `go-pagerduty` dependency in this same PR please.

jedelson-pagerduty:
@imjaroiswebdev should be all set now.

imjaroiswebdev (approved):
Thank you @jedelson-pagerduty We're ready to go now 🎉 💪🏽

PR #722, opened 2023-08-04
document event orchestration attribute type for `integration`

margueritepd:


website/docs/r/event_orchestration.html.markdown
```diff
@@ -37,9 +37,9 @@ The following arguments are supported:
 The following attributes are exported:
 
 * `id` - The ID of the Event Orchestration.
-* `integration` - An integration for the Event Orchestration.
+* `integration` - A list of integrations for the Event Orchestration.
   * `id` - ID of the integration
-  * `parameters`
+  * `parameters` - A list of parameters
     * `routing_key` - Routing key that routes to this Orchestration.
     * `type` - Type of the routing key. `global` is the default type.
```
margueritepd, line 44: In the [terraform registry docs](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/data-sources/event_orchestration), the last two bullets are not displaying as being nested under `parameters`. Any idea how to fix?

website/docs/d/event_orchestrations.html.markdown
```diff
@@ -37,9 +37,9 @@ The following arguments are supported:
 * `name_filter` - The list of the Event Orchestrations which name match `name_filter` argument.
```
margueritepd, line 37: I don't see this in the published docs https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/data-sources/event_orchestration, nor do I see `name_filter` in  I'm not sure why it says `name_filter` in the [code definition](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/master/pagerduty/data_source_pagerduty_event_orchestration.go) for the event orchestration data source so I'm not sure why this markdown file says `name_filter`. Maybe I'm editing the wrong markdown file?

alenapan: Yeah, this is a bit confusing because we have two data sources that look very similar:
- `pagerduty_event_orchestrations`: lets you pull a list of orchestrations matching the given name filter. This is where it's [documented](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/data-sources/event_orchestrations) and [defined in the code](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/master/pagerduty/data_source_pagerduty_event_orchestrations.go)
- `pagerduty_event_orchestration`: returns a single orchestration matching the given name. It's documented and defined in the links that you posted.

website/docs/r/event_orchestration.html.markdown
alenapan: Thank you for fixing this! Hmm not sure why it's not nested properly, but I think there is a TF doc preview tool, let me see if I can find it.

alenapan: @imjaroiswebdev I'm sorry to be tagging you everywhere today but wasn't it you who showed me a markdown preview tool specifically for Terraform docs? Or am I making things up?

PR #723, opened 2023-08-08
Specify the default country code for contact methods

drastawi:
If the default country code is not specified, the API will default to 1. It is creating a permanent diff if not specified in a provider.


imjaroiswebdev (approved):
Hey @drastawi thank you for this contribution. btw, I took to the liberty of replacing the **Default** You added, by a more convenient and appropriate setting for this case. **Computed = true**.

imjaroiswebdev:
Hey @drastawi https://github.com/PagerDuty/terraform-provider-pagerduty/releases/tag/v3.1.1 shipped ✌🏽  Thanks again! 🎉

PR #768, opened 2023-11-10
Feat/add iw inline steps inputs support

brandonw:
This update brings support for using `inline_steps_input`s when building Incident Workflows. `inline_steps_input`s are specialized inputs of certain Actions that can be identified by looking for `@metadata.format="inlineSteps".

These specialized inputs diverge from the standard value format of `string`, and instead are composed of a specific schema that allows inline steps to be configured. This allows for a series of inline steps to be executed conditionally, or within a loop, while still being defined as part of a single workflow.

I have also included several fixes for configured workflows with inputs that have default values, or are ordered differently from the canonical action input ordering. These changes will prevent plan diffs from always existing, even after having just applied the current config.

Test cases were extended for both IncidentWorkflow and IncidentWorkflowTrigger:

```sh
$ PAGERDUTY_ACC_INCIDENT_WORKFLOWS=1 make testacc TESTARGS="-run IncidentWorkflow"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run IncidentWorkflow -timeout 120m
?   	github.com/terraform-providers/terraform-provider-pagerduty	[no test files]
=== RUN   TestAccDataSourcePagerDutyIncidentWorkflow
--- PASS: TestAccDataSourcePagerDutyIncidentWorkflow (8.33s)
=== RUN   TestAccDataSourcePagerDutyIncidentWorkflow_Missing
--- PASS: TestAccDataSourcePagerDutyIncidentWorkflow_Missing (1.51s)
=== RUN   TestAccPagerDutyIncidentWorkflow_import
--- PASS: TestAccPagerDutyIncidentWorkflow_import (6.61s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_import
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_import (15.93s)
=== RUN   TestAccPagerDutyIncidentWorkflow_Basic
--- PASS: TestAccPagerDutyIncidentWorkflow_Basic (10.48s)
=== RUN   TestAccPagerDutyIncidentWorkflow_Team
--- PASS: TestAccPagerDutyIncidentWorkflow_Team (7.98s)
=== RUN   TestAccPagerDutyIncidentWorkflow_InlineInputs
--- PASS: TestAccPagerDutyIncidentWorkflow_InlineInputs (10.30s)
=== RUN   TestFlattenIncidentWorkflowStepsOneGenerated
--- PASS: TestFlattenIncidentWorkflowStepsOneGenerated (0.00s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BadType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BadType (1.10s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithManualType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithManualType (1.09s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ConditionalTypeWithoutCondition
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ConditionalTypeWithoutCondition (1.09s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_SubscribedToAllWithInvalidServices
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_SubscribedToAllWithInvalidServices (1.08s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BasicManual
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BasicManual (13.52s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BasicConditionalAllServices
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BasicConditionalAllServices (9.97s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_CannotChangeType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_CannotChangeType (13.27s)
PASS
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	102.852s
```

## Merging depends on https://github.com/heimweh/go-pagerduty/pull/141

### DIFF
--- a/pagerduty/resource_pagerduty_incident_workflow.go
+++ b/pagerduty/resource_pagerduty_incident_workflow.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -13,6 +14,34 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/heimweh/go-pagerduty/pagerduty"
 )
+
+// This integer controls the level of inline_steps_inputs recursion allowed in the Incident Workflow schema.
+// A value of 1 indicates that the top level workflow steps may specify inline_steps_inputs of an action whose
+// configuration requires it, but a step that exists within an inline_steps_input cannot itself specify an
+// inline_steps_input. For example:
+//
+//	resource "pagerduty_incident_workflow" "test_workflow" {
+//	  name = "Test Terraform Incident Workflow"
+//	  step {
+//	    name   = "Step 1"
+//	    action = "pagerduty.com:incident-workflows:action:1"
+//	    inline_steps_input { # Allowed iff inlineStepSchemaLevel>0
+//	      name  = "Actions"
+//	      step {
+//	        name = "Inline Step 1"
+//	        action = "pagerduty.com:incident-workflows:action:1"
+//	        inline_steps_input { # Allowed iff inlineStepSchemaLevel>1
+//	          ...
+//	            inline_steps_input { # Allowed iff inlineStepSchemaLevel>2
+//	          ...
+//	        }
+//	      }
+//	    }
+//	  }
+//	}
+//
+// Each time the value is incremented by 1, it will allow another level of "inline_steps_input".
+const inlineStepSchemaLevel = 1
 
 func resourcePagerDutyIncidentWorkflow() *schema.Resource {
 	return &schema.Resource{
@@ -77,6 +106,114 @@ func resourcePagerDutyIncidentWorkflow() *schema.Resource {
 								},
 							},
 						},
+						"inline_steps_input": inlineStepSchema(inlineStepSchemaLevel),
+					},
+				},
+			},
+		},
+	}
+}
+
+func inlineStepSchema(level int) *schema.Schema {
+	if level > 1 {
+		return &schema.Schema{
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"name": {
+						Type:     schema.TypeString,
+						Required: true,
+					},
+					"step": {
+						Type:     schema.TypeList,
+						Optional: true,
+						Computed: true,
+						Elem: &schema.Resource{
+							Schema: map[string]*schema.Schema{
+								"name": {
+									Type:     schema.TypeString,
+									Required: true,
+								},
+								"action": {
+									Type:     schema.TypeString,
+									Required: true,
+								},
+								"input": {
+									Type:     schema.TypeList,
+									Optional: true,
+									Computed: true,
+									Elem: &schema.Resource{
+										Schema: map[string]*schema.Schema{
+											"name": {
+												Type:     schema.TypeString,
+												Required: true,
+											},
+											"value": {
+												Type:     schema.TypeString,
+												Required: true,
+											},
+											"generated": {
+												Type:     schema.TypeBool,
+												Computed: true,
+											},
+										},
+									},
+								},
+								"inline_steps_input": inlineStepSchema(level - 1),
+							},
+						},
+					},
+				},
+			},
+		}
+	}
+
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"name": {
+					Type:     schema.TypeString,
+					Required: true,
+				},
+				"step": {
+					Type:     schema.TypeList,
+					Optional: true,
+					Computed: true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"name": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"action": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"input": {
+								Type:     schema.TypeList,
+								Optional: true,
+								Computed: true,
+								Elem: &schema.Resource{
+									Schema: map[string]*schema.Schema{
+										"name": {
+											Type:     schema.TypeString,
+											Required: true,
+										},
+										"value": {
+											Type:     schema.TypeString,
+											Required: true,
+										},
+										"generated": {
+											Type:     schema.TypeBool,
+											Computed: true,
+										},
+									},
+								},
+							},
+						},
 					},
 				},
 			},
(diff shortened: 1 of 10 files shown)


pagerduty/resource_pagerduty_incident_workflow_trigger.go
```diff
@@ -128,6 +128,11 @@ func resourcePagerDutyIncidentWorkflowTriggerDelete(ctx context.Context, d *sche
 }
 
 func validateIncidentWorkflowTrigger(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	oldType, newType := d.GetChange("type")
+	if oldType != nil && oldType != "" && oldType != newType {
+		return fmt.Errorf("trigger type cannot be changed")
+	}
+
```
brandonw, line 135: Did a quick spot check of the trigger stuff while I was here and saw that the config allowed you to change the type, but that change is ignored by the server. Instead of having a diff constantly appear and never be applied, we reject changing the type after the trigger has been created.

pagerduty/resource_pagerduty_incident_workflow.go
```diff
@@ -14,6 +15,8 @@ import (
 	"github.com/heimweh/go-pagerduty/pagerduty"
 )
 
+const inlineStepSchemaLevel = 1
```
brandonw, line 18: Controls how many levels of inline steps we allow. Currently only 1 level is allowed, ie an inline step may not use an action that itself uses an inline step.

pagerduty/resource_pagerduty_incident_workflow.go
```diff
@@ -89,35 +200,95 @@ func resourcePagerDutyIncidentWorkflow() *schema.Resource {
 // These inputs get persisted in the state with `generated` set to `true`
 // but then should not be removed by a `terraform apply`
 func customizeIncidentWorkflowDiff() schema.CustomizeDiffFunc {
-	inputCountRegex := regexp.MustCompile(`^(step\.(\d+)\.input)\.#$`)
```
brandonw, line 92: I reworked this diff to catch a lot of uncaught diff issues originating from inputs with default values (aka generated) as well as ordering issues. The diff will now use the ordering of inputs from the tfstate, but the values from the config. Numbered comments have been added to make the flow a bit more understandable.

brandonw: Of note is that we now match any input change at any level, not just quantity changes at the top level. This means the "calculate new inputs" logic is now recursive, as it takes in a config path, recurses to the correct level, and then deals with the list of inputs at that level.

pagerduty/resource_pagerduty_incident_workflow.go
```diff
@@ -160,7 +331,7 @@ func resourcePagerDutyIncidentWorkflowCreate(ctx context.Context, d *schema.Reso
 		return diag.FromErr(err)
 	}
 
-	iw, err := buildIncidentWorkflowStruct(d)
+	iw, specifiedSteps, err := buildIncidentWorkflowStruct(d)
```
brandonw, line 334: We now calculate the specified inputs while building the struct to send in the request. This makes it easier, as we already recurse through inline inputs while generating the request body, and we already have access to the config values to determine which are generated or not.

pagerduty/resource_pagerduty_incident_workflow.go
```diff
@@ -172,13 +343,7 @@ func resourcePagerDutyIncidentWorkflowCreate(ctx context.Context, d *schema.Reso
...
 	}
 
-	stepIdMapping := map[int]string{}
-	for i, s := range createdWorkflow.Steps {
-		stepIdMapping[i] = s.ID
-	}
-
-	nonGeneratedInputNames := createNonGeneratedInputNamesFromWorkflowStepsWithoutIDs(iw, stepIdMapping)
-	err = flattenIncidentWorkflow(d, createdWorkflow, true, nonGeneratedInputNames)
+	err = flattenIncidentWorkflow(d, createdWorkflow, true, specifiedSteps)
```
brandonw, line 346: When we flatten the workflow after receiving the response, we now use the specified inputs tracked earlier to determine which inputs need `generated=true` or not.

pagerduty/resource_pagerduty_incident_workflow.go
```diff
@@ -322,41 +449,41 @@ func flattenIncidentWorkflow(d *schema.ResourceData, iw *pagerduty.IncidentWorkf
...
+func flattenIncidentWorkflowStepInput(inputs []*pagerduty.IncidentWorkflowActionInput, specifiedInputNames []string) *[]interface{} {
 	newInputs := make([]interface{}, len(inputs))
 
 	for i, v := range inputs {
 		m := make(map[string]interface{})
 		m["name"] = v.Name
 		m["value"] = v.Value
 
-		if !isInputInNonGeneratedInputNames(v, nonGeneratedInputNames) {
+		if !isInputInNonGeneratedInputNames(v, specifiedInputNames) {
```
brandonw, line 486: Once we get to the point where we set generated=true or not, the logic is the same. Now that we pass the specified inputs recursive struct along as we flatten the response, we can easily access what the current specified inputs are.

pagerduty/resource_pagerduty_incident_workflow.go
```diff
@@ -430,7 +626,68 @@ func buildIncidentWorkflowInputsStruct(in interface{}) []*pagerduty.IncidentWork
...
+		}
 		newInputs[i] = &input
 	}
-	return newInputs
+	return newInputs, specifiedInputNames
+}
+
+func buildIncidentWorkflowInlineStepsInputsStruct(in interface{}) (
+	[]*pagerduty.IncidentWorkflowActionInlineStepsInput,
+	map[string][]*SpecifiedStep,
```
brandonw, line 640: These builder functions all now pass a pair of values as their result: the object that is eventually send as part of the request body, and the specified inputs tracking objects that can be referenced as we flatten the response.

pagerduty/resource_pagerduty_incident_workflow_test.go
```diff
@@ -219,46 +441,152 @@ func testAccCheckPagerDutyIncidentWorkflowExists(n string) resource.TestCheckFun
...
+						},
+					},
 				},
 			},
 		},
 	}
-	o := map[string][]string{
-		"abc-123": {"test1-value", "test2-value"},
+	specifiedSteps := []*SpecifiedStep{
+		&SpecifiedStep{
```
imjaroiswebdev, line 518: Could please remove the redundant types declarations on slice elements like `&SpecifiedStep` and `[]*SpecifiedStep` for map values?

imjaroiswebdev (changes requested):
In order for the unit tests to run without issues, please introduce a module replacement pointing to the HEAD of your version of `go-pagerduty` with latest changes.

I left a few comments, which should be addressed first to proceed with merge.

pagerduty/resource_pagerduty_incident_workflow_test.go
brandonw: This is my first time using golang- could you clarify this a bit for me? Is there some automatic magic that happens with slice items that convert structs to a pointer for you, or something like that? So this becomes `specifiedSteps := []SpecifiedStep{ SpecifiedStep{ ... }, ... }` ?

imjaroiswebdev (approved):
Great work @brandonw. Thank you! Inline Inputs enhancement to Incident Workflows will be pretty appreciated by PagerDuty Terraform provider users 🎉 🥳

Issue #782, opened 2023-12-08
documentation pagerduty_service_integration integration_key attribute is deprecated


Marcus-James-Adams:
Hello
#775 introduced the deprecation  notice when running terraform plan & apply

The documentation https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/service_integration states that the attribute has been deprecated, however there is nothing to say what we would be using in it place

eg in the code block below what should we use for the servce_key field?

```
resource "pagerduty_service_integration" "myapp" {
  name    = "my-app-intergration"
  service = pagerduty_service.myapp.id
  vendor  = data.pagerduty_vendor.datadog.id
}

resource "datadog_integration_pagerduty_service_object" "myapp" {
  service_name = pagerduty_service.myapp.name
  service_key  = pagerduty_service_integration.myapp.integration_key
}
```

vperaltac:
The deprecation warning also shows up when using `integration_key` as an output. There is already a PR pending that fixed this issue. Is there an ETA for the fix to be live?

imjaroiswebdev:
This was addressed by https://github.com/PagerDuty/terraform-provider-pagerduty/pull/801. I'd suggest you to upgrade your current PagerDuty TF provider version to **v3.4.1**.

PR #785, opened 2023-12-13
[integration key] do not error with deprecation on output-only usage

drew-u410:
Fixes the `integration_key` deprecation warning when only used on `output` and not `input`.

# Detail
* In https://github.com/PagerDuty/terraform-provider-pagerduty/pull/775, the `integration_key` was marked as `deprecated` with the intent for `input` (makes sense).
* Post-shipping, it was found that this when used on `output` would still provide the `deprecation` message which is not expected or desired (example: https://github.com/PagerDuty/terraform-provider-pagerduty/issues/782).
* By removing `Optional: true`, this should make it `output`-only in Terraform, to my knowledge.

# Testing
* Tests pass, happy to do other things if desired.
```
go test -timeout=30s -parallel=4 github.com/terraform-providers/terraform-provider-pagerduty github.com/terraform-providers/terraform-provider-pagerduty/pagerduty
?   	github.com/terraform-providers/terraform-provider-pagerduty	[no test files]
ok  	github.com/terraform-providers/terraform-provider-pagerduty/pagerduty	4.227s
```

### DIFF
--- a/pagerduty/resource_pagerduty_service_integration.go
+++ b/pagerduty/resource_pagerduty_service_integration.go
@@ -66,7 +66,6 @@ func resourcePagerDutyServiceIntegration() *schema.Resource {
 			},
 			"integration_key": {
 				Type:       schema.TypeString,
-				Optional:   true,
 				Computed:   true,
 				Deprecated: "Assignments or updates to this attribute are not supported by Service Integrations API, it is a read-only value. Input support will be dropped in upcomming major release",
 			},


imjaroiswebdev:
Thanks @drew-u410. The issue was addressed in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/801. However, this and any other contributions from you are very welcomed.

PR #797, opened 2024-01-08
feat: support for sourcing the provider service region from an env var

ricoli:
This would allow setting the service_region provider argument without needing to define it in every instance of the provider. It's of particular importance for customers using the EU region of course.


imjaroiswebdev (approved):
Cool! thanks for contributing with this enhancement @ricoli 🎉 💪🏽

PR #805, opened 2024-01-24
feat: support for sourcing the provider service region from env var (take 2)

ricoli:
Using the correct schema key this time, as opposed to in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/797


imjaroiswebdev (approved):
Cool! This time looks that's going to work @ricoli 💪🏽

Issue #809, opened 2024-01-26
plugin crashes when importing incident workflows


mfmbarros:
Hi there,

Whenever I try to import an incident workflow, either via terraform cli or import blocks, a runtime error occurs in Pagerduty plugin.

### Terraform Version
tested versions: 1.5.7 , 1.7.1

### Affected Resource(s)
- pagerduty_incident_workflow

### Terraform Configuration Files
```hcl
resource "pagerduty_incident_workflow" "my_slack_incident_workflow" {
  description = "Automatically name, create, and link Slack channels to the incident"
  name        = "Slack Channel"
  step {
    action = "pagerduty.com:incident-workflows:create-slack-channel:3"
    name   = "Create a Slack Channel for an Incident"

    input {
      name  = "Slack Team Name"
      value = var.slack_workspace_id
    }
    input {
      name  = "Desired Channel Name"
      value = "my-incident-{{incident.priority}}-{{incident.service.name}}-{{incident.incident_number}}"
    }
  }
}

import {
  to = pagerduty_incident_workflow.my_slack_incident_workflow
  id = "P5***KT"
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
https://gist.github.com/mfmbarros/2084baf4ad8ea37041ec996e5b323e74

### Expected Behavior
Incident workflow imported with success.

### Actual Behavior
Plugin crashes

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. create an incident workflow manually
2. Import incident workflow with terraform plan or terraform import pagerduty_incident_workflow.my_slack_incident_workflow P5***KT
3. Crash with panic output 

### References
- https://github.com/PagerDuty/terraform-provider-pagerduty/issues/786

viktor-f3:
The same problem here. Trying to migrate from response plays to workflows and the docs are lacking on how to configure steps to use escalation policies, so trying to import and compare the results. 

```
Stack trace from the terraform-provider-pagerduty_v3.7.0 plugin:

panic: runtime error: index out of range [0] with length 0

goroutine 63 [running]:
github.com/terraform-providers/terraform-provider-pagerduty/pagerduty.flattenIncidentWorkflowSteps(0xc0001a04d0, {0x277da70, 0x0, 0x1bace00?})
[...]

Issue #786, opened 2023-12-14
importing pagerduty_incident_workflow crashes


simontabor:
### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

v1.6.5

### Affected Resource(s)
- pagerduty_incident_workflow

### Terraform Configuration Files
```hcl
import {
  id = "P2SET0V"
  to = pagerduty_incident_workflow.major_incident
}

resource "pagerduty_incident_workflow" "major_incident" {
  name        = "Major incident workflow"
  description = "Workflow to handle any major incident"
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
[If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.
](https://gist.github.com/simontabor/e1850f8cda79ab9b55d6907a80144b79)
### Expected Behavior
Resource imported successfully

### Actual Behavior
Crash

### Steps to Reproduce
1. Create an incident workflow in the UI
2. Import the workflow to terraform
3. terraform plan - get a crash

### Important Factoids
Created the incident workflow using the template.
Before publishing the workflow, the plan worked fine (but it detected no `steps` in the existing resource)
After publishing the workflow, the provider crashes

### References
Can't see any

PR #815, opened 2024-02-06
Adding an escape for an empty specifiedSteps slice when calling flattenIncidentWorkflowSteps

jkueh:
I had encountered the same issue as reported here when importing an incident workflow:
* https://github.com/PagerDuty/terraform-provider-pagerduty/issues/786
* https://github.com/PagerDuty/terraform-provider-pagerduty/issues/809

I'm not 100% sure if this handles the resource attributes appropriately, would love a sanity check :)
Testing this locally worked for me, as far as I can tell - Or at least let me finish importing my resources.

I've run the local acceptance tests, but it looks like one test fails before I've added my changes, so I can't be too sure 🤔 
image

### DIFF
--- a/pagerduty/resource_pagerduty_incident_workflow.go
+++ b/pagerduty/resource_pagerduty_incident_workflow.go
@@ -488,6 +488,9 @@ func flattenIncidentWorkflowSteps(iw *pagerduty.IncidentWorkflow, specifiedSteps
 	newSteps := make([]map[string]interface{}, len(iw.Steps))
 	for i, s := range iw.Steps {
 		m := make(map[string]interface{})
+		if len(specifiedSteps) == 0 {
+			continue
+		}
 		specifiedStep := *specifiedSteps[i]
 		m["id"] = s.ID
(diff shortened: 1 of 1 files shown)


mfmbarros:
I tested those changes locally and I confirm the plugin crash is gone when running plan with incident workflow import block on. However, in that plan I have Terraform attempting to update incident workflow by adding the steps present in declared resource but those steps are already created in PD instance so I was expecting an import without modifications rather than with update in the resource.

jkueh:
Thanks! Will give that scenario a test and see if I can work out which bit is misbehaving 👍

imjaroiswebdev:
Hi @jkueh thank you for the time you invested trying to fix this issue. The update you are proposing would save the provider from crashing when importing IW, however, it will also skip the fetching of the steps already set up on PagerDuty environment. A more complete patch for this has been merged in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/820.

This patch for this will be released in a new version of the provider today. Please stay tuned :v:

PR #826, opened 2024-02-27
Address: Time zone validation is not aligned with the API

### LINKED ISSUE #796: Time zone validation is not aligned with the API
The time zone validation that was introduced in https://github.com/PagerDuty/terraform-provider-pagerduty/pull/473 does not seem to be compatible with what is present on the API, for example US/Mountain is valid in the provider checks but it fails on the API. Likely the solution is to align the version of the tz info between the provider and the API.

### Terraform Version
1.6.6

### Affected Resource(s)
- user
- service
- service event rule
- schedule
- ruleset rule

```hcl
timezone = "US/Mountain"
```

### Panic Output
time zone is not a valid tzinfo identifier

### Expected Behavior
Should throw an error

### Actual Behavior
Calls the API

### Steps to Reproduce
1. `terraform apply`

### References
https://stackoverflow.com/questions/64619841/time-loadlocation-gives-unknown-timezone-for-us-pacific-new-timezone

imjaroiswebdev:
Closes #796


cjgajard (approved):

util/util.go
```diff
@@ -276,3 +277,180 @@ func ResourcePagerDutyParseColonCompoundID(id string) (string, string, error) {
...
 	return parts[0], parts[1], nil
 }
+
+func ValidateTZValueDiagFunc(v interface{}, p cty.Path) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	value := v.(string)
+	valid := false
+
+	if sort.SearchStrings(validTZ, value) < len(validTZ) {
```
cjgajard, line 287: sort.Search has a weird interface, we thus need to check if `validTZ[index] == value` in order have a valid match.

Example: https://go.dev/play/p/Osx8HcsVL1F
image

cjgajard (changes requested):

cjgajard (approved):

PR #838, opened 2024-03-14
fix(team membership): Allow 404 to propagate to avoid registering it in disassociated eps list

scohen-nd:
This is related to a bug where when deleting team membership you can get errors when the same user exists in other teams escalation policies (due to the fact that it will try to reattach all the EP's to the team, even when getting 404 error).
TLDR, Line 289 will never get hit if we keep that logic in place.

## Acceptance test results...

```sh
$ make testacc TESTARGS='-run TestAccPagerDutyTeamMembership'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccPagerDutyTeamMembership -timeout 120m
?       github.com/PagerDuty/terraform-provider-pagerduty       [no test files]
=== RUN   TestAccPagerDutyTeamMembership_import
--- PASS: TestAccPagerDutyTeamMembership_import (20.67s)
=== RUN   TestAccPagerDutyTeamMembership_importWithRole
--- PASS: TestAccPagerDutyTeamMembership_importWithRole (19.72s)
=== RUN   TestAccPagerDutyTeamMembership_Basic
--- PASS: TestAccPagerDutyTeamMembership_Basic (18.80s)
=== RUN   TestAccPagerDutyTeamMembership_WithRole
--- PASS: TestAccPagerDutyTeamMembership_WithRole (16.48s)
=== RUN   TestAccPagerDutyTeamMembership_WithRoleConsistentlyAssigned
--- PASS: TestAccPagerDutyTeamMembership_WithRoleConsistentlyAssigned (27.50s)
=== RUN   TestAccPagerDutyTeamMembership_DestroyWithEscalationPolicyDependant
--- PASS: TestAccPagerDutyTeamMembership_DestroyWithEscalationPolicyDependant (27.87s)
PASS
ok      github.com/PagerDuty/terraform-provider-pagerduty/pagerduty     132.341s
testing: warning: no tests to run
PASS
ok      github.com/PagerDuty/terraform-provider-pagerduty/pagerdutyplugin       0.959s [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/PagerDuty/terraform-provider-pagerduty/util  1.196s [no tests to run]
```

### DIFF
--- a/pagerduty/resource_pagerduty_team_membership.go
+++ b/pagerduty/resource_pagerduty_team_membership.go
@@ -277,9 +277,12 @@ func dissociateEPsFromTeam(c *pagerduty.Client, teamID string, eps []string) ([]
 	for _, ep := range eps {
 		retryErr := retry.Retry(2*time.Minute, func() *retry.RetryError {
 			_, err := c.Teams.RemoveEscalationPolicy(teamID, ep)
-			if err != nil {
+			if err != nil && !isErrCode(err, 404) {
 				time.Sleep(2 * time.Second)
 				return retry.RetryableError(err)
+			}
+			if isErrCode(err, 404) {
+				return retry.NonRetryableError(err)
 			}
 			return nil
 		})
--- a/pagerduty/resource_pagerduty_team_membership.go
+++ b/pagerduty/resource_pagerduty_team_membership.go
@@ -277,7 +277,7 @@ func dissociateEPsFromTeam(c *pagerduty.Client, teamID string, eps []string) ([]
 	for _, ep := range eps {
 		retryErr := retry.Retry(2*time.Minute, func() *retry.RetryError {
 			_, err := c.Teams.RemoveEscalationPolicy(teamID, ep)
-			if err != nil && !isErrCode(err, 404) {
+			if err != nil {
 				time.Sleep(2 * time.Second)
 				return retry.RetryableError(err)
 			}
--- a/pagerduty/resource_pagerduty_team_membership.go
+++ b/pagerduty/resource_pagerduty_team_membership.go
@@ -281,7 +281,7 @@ func dissociateEPsFromTeam(c *pagerduty.Client, teamID string, eps []string) ([]
 				time.Sleep(2 * time.Second)
 				return retry.RetryableError(err)
 			}
-			if isErrCode(err, 404) {
+			if err != nil && isErrCode(err, 404) {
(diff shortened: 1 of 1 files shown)


pagerduty/resource_pagerduty_team_membership.go
```diff
@@ -277,7 +277,7 @@ func dissociateEPsFromTeam(c *pagerduty.Client, teamID string, eps []string) ([]
 	for _, ep := range eps {
 		retryErr := retry.Retry(2*time.Minute, func() *retry.RetryError {
 			_, err := c.Teams.RemoveEscalationPolicy(teamID, ep)
-			if err != nil && !isErrCode(err, 404) {
+			if err != nil {
```
cjgajard, line 280: Nice catch! But I think the original intention of `&& !isErrCode(err, 404)` is to skip retry times. Thus, we should return a `retry.RetryableError(err)` if `isErrCode(err, 404)` inside of this error handling.

scohen-nd: You certain it should not lead to calling continue at line 292? I think the 404 is exactly for that escalation policy not found - hence skipping it.
The current bug I experience in my environment is that since it is not calling continue - it adds the non relevant/existent escalation policy to the epsDissociatedFromTeam in line 295 and than breaks when it tried to add it back.

cjgajard: I'm sorry, I meant `NonRetryableError`.
You're are right, and we're intending the same. Now (before your changes) the subfunction passed to retry.Retry returns `nil` when `err` is a 404. With your changes it will sleep and return a retryable error. I want to send the execution directly into 292 without waiting 2 minutes.

```
| case          | err == nil | err is 404                   | err is anything else      |
| ------------- | ---------- | ---------------------------- | ------------------------- |
| now           | nil        | nil                          | retry.RetryableError(err) |
| your changes  | nil        | retry.RetryableError(err)    | retry.RetryableError(err) |
| my suggestion | nil        | retry.NonRetryableError(err) | retry.RetryableError(err) |
```

scohen-nd: Ahh, make sense, adjusted the solution accordingly.

cjgajard (approved):

imjaroiswebdev (approved):
Hey @scohen-nd good catch! Thanks for improving this 💪🏽 🎉

PR #847, opened 2024-04-08
Delete a non-actionable warning on the `pagerduty_event_orchestration_integration` resource

### LINKED ISSUE #704: Noisy and unnecessary warning message when using `pagerduty_event_orchestration_integration`
Hi,

currently when creating an event orchestration integration resource like:
```tf
resource "pagerduty_event_orchestration_integration" "integration" {
  event_orchestration = pagerduty_event_orchestration.event_orchestration.id
  label = "Example integration"
}
```

Every time I ran `terraform preview` it will print a warning message like:
```
Modifying the event_orchestration property of the 'pagerduty_event_orchestration_integration' resource will cause all future events sent with this integration's routing key to be evaluated against the new Event Orchestration.
```
Even though I didn't make any modification to that resource.

This appears to happen because the code here https://github.com/PagerDuty/terraform-provider-pagerduty/blob/3b824ee4c23eae91849f8890fbc8931fc6959271/pagerduty/resource_pagerduty_event_orchestration_integration.go#L64-L68 appends this warning message unconditionally and in call cases.

Could you please fix this and only print this warning message if there was actually a change in the resource?

This happens on the latest tf provider version 2.15.0 but has been there in previous versions too.

### LINKED ISSUE #839: Disable warning "Warning: Modifying the event_orchestration property of the 'pagerduty_event_orchestration_integration' resource will cause all future events sent with this integration's routing key to be evaluated against the new Event Orchestration."
### Terraform Version
Terraform v1.1.4
on darwin_arm64
+ provider registry.terraform.io/hashicorp/random v3.6.0
+ provider registry.terraform.io/pagerduty/pagerduty v3.10.0

### Affected Resource(s)
pagerduty_event_orchestration_integration

```hcl
resource "pagerduty_event_orchestration_integration" "integration" {
  for_each            = tomap({ for i in var.event_orcastrations : i.event_orcastration_name => i })
  event_orchestration = pagerduty_event_orchestration.event_orchestration[each.value.event_orcastration_name].id
  label = "${each.value.event_orcastration_name}"
}
```

Do I need to see this every time I plan? 

```
Warning: Modifying the event_orchestration property of the 'pagerduty_event_orchestration_integration' resource will cause all future events sent with this integration's routing key to be evaluated against the new Event Orchestration.

│   with module.test.pagerduty_event_orchestration_integration.integration["some key"],
│   on ../pd.tf line 10, in resource "pagerduty_event_orchestration_integration" "integration":
│   10:   event_orchestration = pagerduty_event_orchestration.event_orchestration[each.value.event_orcastration_name].id
```

Am I doing something wrong or bad? Can I tell PD I know what I'm doing? This warning does a lot to obfuscate actual output and I'd like to turn it off (without disabling ALL tf warnings). Unless there is something wrong I'm doing here.

alenapan:
### Context
When we just launched multiple EO Integrations we wanted to emphasize that modifying the `event_orchestration` attribute will result in an ingestion-level change where all future events sent to this integration will now be evaluated against a new orchestration. But since this warning is non-actionable and we also have this information in the Provider docs, we decided that the warning is rather an annoyance than a helpful message. This PR deletes the warning and as a result fixes the following issues:
- Closes #704 
- Closes #839

### DIFF
--- a/pagerduty/resource_pagerduty_event_orchestration_integration.go
+++ b/pagerduty/resource_pagerduty_event_orchestration_integration.go
@@ -28,7 +28,6 @@ func resourcePagerDutyEventOrchestrationIntegration() *schema.Resource {
 			"event_orchestration": {
 				Type:             schema.TypeString,
 				Required:         true,
-				ValidateDiagFunc: addIntegrationMigrationWarning(),
 			},
 			"id": {
 				Type:     schema.TypeString,
@@ -55,20 +54,6 @@ func resourcePagerDutyEventOrchestrationIntegration() *schema.Resource {
 				},
 			},
 		},
-	}
-}
-
-func addIntegrationMigrationWarning() schema.SchemaValidateDiagFunc {
-	return func(v interface{}, p cty.Path) diag.Diagnostics {
-		var diags diag.Diagnostics
-
-		diags = append(diags, diag.Diagnostic{
-			Severity:      diag.Warning,
-			Summary:       "Modifying the event_orchestration property of the 'pagerduty_event_orchestration_integration' resource will cause all future events sent with this integration's routing key to be evaluated against the new Event Orchestration.",
-			AttributePath: p,
-		})
-
-		return diags
 	}
 }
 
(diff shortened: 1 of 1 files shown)


imjaroiswebdev (approved):
Thank you @alenapan, based on the feedback about for this warning, this update should be well received 👍🏽

PR #881, opened 2024-06-12
Add the option to ignore TLS certificate errors when calling the PD API.

dazoakley:
Hi folks,

Thanks for the awesome module.

This is my first attempt at writing terraform provider code (and I'm pretty green to golang) so please be gentle. :smile: 

This update makes it possible to ignore TLS certs if you really really need to, i.e. when you're behind a corporate proxy that messes with with the certificates...

Is this something you'd accept as a change? I've tested it on our CI server and it applies nicely.

I wasn't sure if I should update the changelog or not, please let me know and I can remove that edit if needed.

Thanks!

### DIFF
--- a/pagerduty/config.go
+++ b/pagerduty/config.go
@@ -1,6 +1,7 @@
 package pagerduty
 
 import (
+	"crypto/tls"
 	"fmt"
 	"log"
 	"net/http"
@@ -37,6 +38,9 @@ type Config struct {
 	// UserAgent for API Client
 	UserAgent string
 
+	// Do not verify TLS certs for HTTPS requests - useful if you're behind a corporate proxy
+	InsecureTls bool
+
 	APITokenType *pagerduty.AuthTokenType
 
 	AppOauthScopedTokenParams *persistentconfig.AppOauthScopedTokenParams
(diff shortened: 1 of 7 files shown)


pagerduty/config.go
```diff
@@ -72,7 +76,12 @@ func (c *Config) Client() (*pagerduty.Client, error) {
 	var httpClient *http.Client
 	httpClient = http.DefaultClient
 	httpClient.Timeout = 1 * time.Minute
-	httpClient.Transport = logging.NewTransport("PagerDuty", http.DefaultTransport)
+
```
imjaroiswebdev, line 79: @dazoakley please add this same update to `Config.SlackClient` 🙏🏽

imjaroiswebdev (changes requested):
Hey there @dazoakley! Great addition, only thing missing is to also add the update for `SlackClient` and We should be ready to merge. Additionally, please rebase onto `master`. Thanks!

pagerduty/config.go
dazoakley: Hi @imjaroiswebdev, thanks for the review! No probs at all updated and rebased. :+1:

imjaroiswebdev: Thank you!

PR #912, opened 2024-07-20
Fix service_integration documentation

viasnake:
Fix service_integration documentation.

`body_mode`, `from_email_mode`, and `subject_mode` of the Email filters can specified with the `no-match` option.
However, the documentation states that only `always` and `match` can be specified.

https://github.com/PagerDuty/terraform-provider-pagerduty/blob/4741e4139fae3e7cd9a80986a064959dd476618c/pagerduty/resource_pagerduty_service_integration.go#L277-L311


cjgajard (approved):
Thank you @viasnake !

PR #937, opened 2024-10-09
Try to ensure escalation rule targets are added in same order as the plan

cjgajard:


imjaroiswebdev (approved):
LGTM! Thank you @cjgajard 🥇

PR #939, opened 2024-10-15
Fix `pagerudty.com` typo in docs

thoward:


cjgajard (approved):

thoward:
Thanks @cjgajard. Do you mind approving the workflow for this PR as well? I think because I'm a first-time contributor it requires approval to run the automation.

PR #941, opened 2024-10-22
Install a fresh terraform binary if we're unable to locate a matching version in PATH.

gdborton:
So full disclosure... I don't know go. I stumbled my way through this and managed to run a simple test w/ vscode.

I'm not sure if this is the best approach to fix the problem that I'm seeing, but I want to try to move along the fix as much as I can.

This is meant to address an issue that we're seeing atm in Terraform Cloud. We get this message when trying to apply our plan:
> terraform binary version >= 1.0.6 not found in $PATH

I looked through the commit history and spotted that this was changed in March to not download a fresh binary. I tried reverting our usage to the version that pulls a fresh binary, but then hit this message:
image

According to this [SO post](https://stackoverflow.com/questions/70763407/blocks-of-type-cloud-are-not-expected-here-for-integrating-with-terraform-clou) it seems to be a requirement of 1.1 in Terraform Cloud while the older version of this provider installs 1.0.6.

I took a screenshot of my poorly written test output (since deleted as I don't know go and couldn't write the mocks):
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 ## Requirements
 
--	[Terraform](https://www.terraform.io/downloads.html) 0.12.x
+-	[Terraform](https://www.terraform.io/downloads.html) >= 1.1.0
 -	[Go](https://golang.org/doc/install) 1.20 (to build the provider plugin)
 
 ## Building the Provider
--- a/pagerduty/terraform_state_snapshot_helper.go
+++ b/pagerduty/terraform_state_snapshot_helper.go
@@ -3,6 +3,7 @@ package pagerduty
 import (
 	"context"
 	"fmt"
+	"log"
 	"os"
 	"strings"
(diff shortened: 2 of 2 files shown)


gdborton:
Following up here as we managed to get past this issue. It seems that this helper file is only invoked if you're deleting a team that is used in an escalation policy with only a single layer. In our case we were "renaming" the team, but this was being applied as a delete/recreate.

We were able to unblock by manually adding an additional layer to the blocking escalation policy. TF was then able to do it's thing.

I don't know if the logic in this PR exactly needs to land, _but_ there definitely is a bug today in the PagerDuty TF provider than surfaces in TF Cloud when you invoke this helper.

cjgajard:
Thank you for your contribution @gdborton ! This surely was difficult to catch! Good job💪!

PR #945, opened 2024-10-29
Aa 1837

juliannagreen1:
Adds new field `only_invocable_on_unresolved_incidents` to Automation Action's schema.


mtrudel (approved):
LGTM though I'm no expert

pagerduty/resource_pagerduty_automation_actions_action_test.go
```diff
@@ -214,7 +218,8 @@ resource "pagerduty_automation_actions_action" "foo" {
 		process_automation_job_id = "pa_job_id_123"
 		process_automation_job_arguments = "-arg 1"
 		process_automation_node_filter = "tags: production"
-	  }
+	  },
```
cjgajard, line 221: The trailing comma is breaking the test

pagerduty/data_source_pagerduty_automation_actions_action_test.go
```diff
@@ -75,6 +75,7 @@ resource "pagerduty_automation_actions_action" "test" {
 		process_automation_job_arguments = "-arg 1"
 		process_automation_node_filter = "tags: production"
 	  }
+	only_invocable_on_unresolved_incidents: true
```
cjgajard, line 78: use `=` instead of `:`

pagerduty/resource_pagerduty_automation_actions_action.go
```diff
@@ -298,6 +308,10 @@ func resourcePagerDutyAutomationActionsActionRead(d *schema.ResourceData, meta i
 			if automationActionsAction.ActionClassification != nil {
 				d.Set("action_classification", &automationActionsAction.ActionClassification)
 			}
+
+			if automationActionsAction.OnlyInvocableOnUnresolvedIncidents != nil {
+				d.Set("only_invocable_on_unresolved_incidents", &automationActionsAction.OnlyInvocableOnUnresolvedIncidents)
```
cjgajard, line 313: As `automationActionsAction.OnlyInvocableOnUnresolvedIncidents` is a pointer to bool the value should be its dereference (i.e. use `*` instead of `&`).

pagerduty/resource_pagerduty_automation_actions_action_test.go
```diff
@@ -84,6 +85,7 @@ func TestAccPagerDutyAutomationActionsActionTypeProcessAutomation_Basic(t *testi
 					resource.TestCheckResourceAttrSet("pagerduty_automation_actions_action.foo", "modify_time"),
 					resource.TestCheckResourceAttrSet("pagerduty_automation_actions_action.foo", "runner_id"),
 					resource.TestCheckResourceAttr("pagerduty_automation_actions_action.foo", "runner_type", "runbook"),
+					resource.TestCheckResourceAttr("pagerduty_automation_actions_action.foo", "only_invocable_on_unresolved_incidents", "true"),
```
cjgajard, line 88: please, test the "false" value here setting explicitly the value in `testAccCheckPagerDutyAutomationActionsActionTypeProcessAutomationConfigUpdated`

pagerduty/resource_pagerduty_automation_actions_action.go
```diff
@@ -153,6 +158,11 @@ func buildAutomationActionsActionStruct(d *schema.ResourceData) (*pagerduty.Auto
 		automationActionsAction.ModifyTime = &val
 	}
 
+	if attr, ok := d.GetOk("only_invocable_on_unresolved_incidents"); ok {
```
cjgajard, line 161: `GetOk` will return `ok` equal to `false` even when the value is explicitly set to `false`. Since the client defines this field as `omitempty` it would never get changed to false.
I'd change this whole `if` block to:
```
attr, _ :=  d.Get("only_invocable_on_unresolved_incidents").(bool)
automationActionsAction.OnlyInvocableOnUnresolvedIncidents = &attr
```

PR #962, opened 2025-01-07
Add incident types resources

cjgajard:
* data pagerduty_incident_type
* data pagerduty_incident_type_custom_field
* resource pagerduty_incident_type
* resource pagerduty_incident_type_custom_field


website/docs/r/incident_type.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+page_title: "PagerDuty: pagerduty_incident_type"
+sidebar_current: "docs-pagerduty-resource-incident_type"
+description: |-
+  Creates and manages a incident_type in PagerDuty.
+---
+
+# pagerduty\_incident\_type
+
+An [incident\_type](https://developer.pagerduty.com/api-reference/1981087c1914c-create-an-incident-type)
+are a feature which will allow customers to categorize incidents, such as a
```
akosasante, line 12:
```suggestion
is a feature which allows customers to categorize incidents, such as a
```

website/docs/r/incident_type.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+  description = "Internal incidents not facing customer"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the Incident Type. Usage of the suffix `_default` is prohibited. This cannot be changed once the incident type has been created.
+* `display_name` - (Required) The display name of the Incident Type. Usage of the prefix PD, PagerDuty, Default is prohibited.
```
akosasante, line 36:
```suggestion
* `display_name` - (Required) The display name of the Incident Type. Usage of the prefixes PD, PagerDuty, or the suffixes Default, or (Default) is prohibited.
```

website/docs/r/incident_type.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+* `id`  - The unique identifier of the incident type.
+* `type`  - A string that determines the schema of the object.
+
+## Import
+
+Services can be imported using the `id`, e.g.
+
+```
+$ terraform import pagerduty_service.main P12345
```
akosasante, line 51: Should this be talking about types, rather than services?

website/docs/r/incident_type_custom_field.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+  Creates and manages a incident type custom field in PagerDuty.
+---
+
+# pagerduty\_incident\_type\_custom\_field
+
+An [incident type custom fields](https://developer.pagerduty.com/api-reference/423b6701f3f1b-create-a-custom-field-for-an-incident-type)
+are a feature which will allow customers to extend Incidents with their own
+custom data, to provide additional context and support features such as
+customized filtering, search and analytics. Custom Fields can be applied to
+different incident types.
```
akosasante, line 15:
```suggestion
different incident types. Child types will inherit custom fields from their parent types.
```

akosasante (commented):
Tested:

- [x] able to create an incident type (though left a comment, seems like it creates it disabled by default)
- [x] able to update an incident type
- [x] able to disable an incident type
- [x] able to import an existing type
- [x] unable to delete a type (as expected; and a warning is printed)
- [ ] able to use `data` block to refer to an existing type by Display Name
- [ ]  able to create a field under a given type
- [ ] able to update a field
- [ ] able to disable a field
- [ ] able to delete a field
- [ ] able to import an existing field
- [ ] able to refer to an existing field with data block
- [ ] able to handle expected errors/validation errors and print (with the caveat that for some validation errors, it fails to print because we haven't yet fixed the nested errors on our end: https://pagerduty.atlassian.net/browse/INCIDENTS-2186)

website/docs/r/incident_type_custom_field.html.markdown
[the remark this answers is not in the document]
cjgajard: Yes, I opted into not implementing a `incident_type_custom_field_option` resource nor datasource because [they don't have a name](https://developer.pagerduty.com/api-reference/7ac20fa48535e-get-a-field-option-on-a-custom-field), hence they can't be searched by anything other than value (or ID).

Also, the requirement for a `field_options` in `*_fixed` fields made it difficult to use with an option resource.
If you create `field_options = ["a", "b"]` and then create a `resource "incident_type_custom_field_option" "c" { value = "c" }` it would break its parent definition.

Now they're accesible through their parent field `field_options`.

website/docs/r/incident_type.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the Incident Type. Usage of the suffix `_default` is prohibited. This cannot be changed once the incident type has been created.
+* `display_name` - (Required) The display name of the Incident Type. Usage of the prefix PD, PagerDuty, Default is prohibited.
+* `parent_type` - (Required) The parent type of the Incident Type. Either name or id of the parent type can be used.
+* `description` - A succinct description of the Incident Type.
+* `enabled`  - State of this Incident Type object. Defaults to true if not provided.
```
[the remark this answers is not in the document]
cjgajard: I was sending an explicit `"enabled": false` when it was not provided. Fixed to match the documentation.

website/docs/r/incident_type_custom_field.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+  field_type   = "multi_value"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) [Updating causes resource replacement] The name of the custom field.
+* `incident_type` - (Required) [Updating causes resource replacement] The id of the incident type the custom field is associated with.
```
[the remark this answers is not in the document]
cjgajard: Thank you for catching this, yes, it should've been `pagerduty_incident_type_custom_field`.

akosasante (approved):
Re-tested and incident type was created in enabled mode when that property is not provided now :+1: ✅

All feedback has been addressed. Functionality-wise LGTM! (I'm not super familiar with golang so I can't speak to that end of things but it all works :D)

go.mod
```diff
@@ -76,3 +76,5 @@ require (
 	google.golang.org/protobuf v1.33.0 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 )
+
+replace github.com/PagerDuty/go-pagerduty => github.com/cjgajard/go-pagerduty v0.0.0-20250107134752-8a1269887cc8
```
imjaroiswebdev, line 80: Please remember to drop this replace and update deps after https://github.com/PagerDuty/go-pagerduty/pull/546 gets merged.

imjaroiswebdev (changes requested):

Issue #840, opened 2024-04-03
Error creating Incident Workflow actions that have no inputs


gsreynolds:
The Create Zoom Meeting action has no inputs, but the Terraform provider doesn't support that currently:
```
resource "pagerduty_incident_workflow" "major_incident" {
  description = "Major Incident
  name        = "Major Incident"
  team        = null
  step {
    action = "pagerduty.com:incident-workflows:create-zoom-meeting:4"
    name   = "Create a Zoom Meeting"
  }
}
```
results in an error.

So nonexistent inputs are required to successfully create the Incident Workflow
```
resource "pagerduty_incident_workflow" "major_incident" {
  description = "Major Incident
  name        = "Major Incident"
  team        = null
  step {
    action = "pagerduty.com:incident-workflows:create-zoom-meeting:4"
    name   = "Create a Zoom Meeting"
    input {
      name  = "Not Required"
      value = "Not Required"
    }
  }
}
```

however, since that input doesn't exist, that workaround introduces a perma-diff as although the action will be created, the fake input will not.

Ideally the steps should accept zero or more inputs, not one or more as currently.

PR #992, opened 2025-03-13
Don't omit 'inputs' field in pagerduty_incident_workflow.step as it's required by the API

tganderson0:
Currently, the provider omits fields that are empty when serializing to JSON. However, the API endpoint `https://api.pagerduty.com/incident_workflows/id`, which is used for updating Incident Workflows, requires that the field `inputs` exists in the payload. This makes it so that you cannot have the create zoom room action defined in terraform, without a 'fake' input which causes perpetual diffs.

This addresses the issue in https://github.com/PagerDuty/terraform-provider-pagerduty/issues/840.

**API Response without the inputs defined**
```py
import requests

url = "https://api.pagerduty.com/incident_workflows/id"

payload = { "incident_workflow": {
        "name": "Example Incident Workflow",
        "description": "This Incident Workflow is an example",
        "steps": [
            {
                "name": "Send Status Update",
                "action_configuration": {
                    "action_id": "pagerduty.com:incident-workflows:send-status-update:1",
                }
            }
        ]
    } }
headers = {
    "Accept": "application/json",
    "Content-Type": "application/json",
    "Authorization": "Token token=---------"
}

response = requests.put(url, json=payload, headers=headers)

print(response.json())

> {'message': 'Invalid body: "incident_workflow.steps[0].action_configuration.inputs" is required'}
```

### DIFF
--- a/vendor/github.com/heimweh/go-pagerduty/pagerduty/incident_workflow.go
+++ b/vendor/github.com/heimweh/go-pagerduty/pagerduty/incident_workflow.go
@@ -34,7 +34,7 @@ type IncidentWorkflowStep struct {
 type IncidentWorkflowActionConfiguration struct {
 	ActionID          string                                    `json:"action_id,omitempty"`
 	Description       *string                                   `json:"description,omitempty"`
-	Inputs            []*IncidentWorkflowActionInput            `json:"inputs,omitempty"`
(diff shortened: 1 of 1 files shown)


cjgajard:
Thank you for finding this and contributing a PR! But in order to prevent conflicts on vendoring, can you please redo this PR in the source repo? https://github.com/heimweh/go-pagerduty

Issue #999, opened 2025-03-27
Perpetual diff for pagerduty_slack_connection.config.events


tmccombs:
It appears that `pagerduty_slack_connection.config.events` is treated as a list, but it should really be a set instead. Pagerduty appears to use some kind of internal ordering, (or possibly the order is non-deterministic?) when returning the set of events, and if you don't supply those events in the correct order, then every time you run a plan, terraform will say there is a change to change the order from the order pagerduty returns to the order that you have in config. And the order pagerduty returns doesn't appear to be alphabetical, so I can't simply call a `sort()` on the events to get it in the right order.

## What I expect should happen

The resource shouldn't care about the order of the events. Either define the events as a set in the schema, or use a custom diff to ignore changes that only change the order.

PR #1000, opened 2025-03-27
Make event config for slack connections a set

tganderson0:
Updates the `pagerduty_slack_connection` to use `schema.TypeSet` instead of `schema.TypeList`, as the order the events are defined shouldn't report a change in a plan output.

Should resolve https://github.com/PagerDuty/terraform-provider-pagerduty/issues/999.


cjgajard (approved):

cjgajard:
Thank you very much @tganderson0!

Issue #671, opened 2023-04-06
Cannot create incident_workflow_trigger that executes on incident creation


modwyer42:
The implementation of the `pagerduty_incident_workflow_trigger` resource does not fully align to the API and prevents creation of triggers that execute on incident creation.
Web UI example of this trigger definition:
![image](https://user-images.githubusercontent.com/15152045/230275630-9cb99dc1-8f5f-4f6c-9970-4fb4b8390327.png)

As per the [API reference](https://developer.pagerduty.com/api-reference/7d0e8f01960aa-update-a-trigger), in order to create a trigger that executes automatically on incident creation the `condition` value must be set to an empty string (`""`) with type set to `conditional`.
![image](https://user-images.githubusercontent.com/15152045/230275313-45ccd4ad-49ae-4cb9-94cc-18c99c6d30e5.png)

The provider is not handling/nulling this empty string and so thinks that a condition was not provided, leading to it returning the error:
![image](https://user-images.githubusercontent.com/15152045/230275249-8d9530ca-1c6b-466d-b973-deeed59010a3.png)

A direct API POST successfully configures the desired trigger when using the following payload:
```json
{
 "trigger_type": "conditional",
 "workflow": {
  "id": "PZXQRKC"
 },
 "services": [
  {
   "id": "PC9TV29"
  }
 ],
 "condition": ""
}
```

### Terraform Version
Terraform v1.4.2
on linux_amd64

PagerDuty provider version: 2.11.2

### Affected Resource(s)
- pagerduty_incident_workflow_trigger

### Terraform Configuration Files

**_Note:-** Simplified configuration example provided.

```hcl
terraform {
  required_providers {
    pagerduty = {
      source  = "pagerduty/pagerduty"
      version = "~>2.9"
    }
  }
}

resource "pagerduty_incident_workflow" "example" {
  name         = "Example Incident Workflow"
  description  = "This Incident Workflow is an example"
  step {
    name           = "Send Status Update"
    action         = "pagerduty.com:incident-workflows:send-status-update:1"
    input {
      name = "Message"
      value = "Example status message sent on {{current_date}}"
    }
  }
}

resource "pagerduty_incident_workflow_trigger" "automatic_trigger" {
  workflow                   = pagerduty_incident_workflow.example.id
  type                       = "conditional"
  condition                  = ""
  subscribed_to_all_services = true
}
```

### Debug Output
N/A

### Expected Behavior
[...]

PR #1012, opened 2025-05-02
Fix issue where condition cannot currently be an empty string

jamesforwardnwboxed:
This should fix: https://github.com/PagerDuty/terraform-provider-pagerduty/issues/671

An empty string is perfectly valid for `condition` where type is set to `conditional` - this effectively sets it to trigger on incident creation. The issue is, `GetOk` isn't sufficient for this, because an error will return if it's the zero value (`""`) despite the fact we actually want this.

Specifying `"incident.status matches 'triggered'"` as a workaround has the unfortunate effect that when the ticket is reassigned, it still counts as 'triggered' rather than created - effectively re-running the workflow if the ticket is reassigned, which may not be desired. Adding this fix solves this issue so we can reliably use the Terraform provider.

### DIFF
--- a/pagerduty/resource_pagerduty_incident_workflow_trigger.go
+++ b/pagerduty/resource_pagerduty_incident_workflow_trigger.go
@@ -262,6 +262,8 @@ func flattenIncidentWorkflowEnabledServices(s []*pagerduty.ServiceReference) []s
 }
 
 func buildIncidentWorkflowTriggerStruct(d *schema.ResourceData, forUpdate bool) (*pagerduty.IncidentWorkflowTrigger, error) {
+	triggerType := d.Get("type").(string)
+
 	iwt := pagerduty.IncidentWorkflowTrigger{
 		SubscribedToAllServices: d.Get("subscribed_to_all_services").(bool),
 	}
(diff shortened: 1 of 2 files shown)


cjgajard:
Thank you for this contribution @jamesforwardnwboxed ! The main reason we did not merge this yet is because it lacks an acceptance test. Do you think you can include one? You can use TestAccPagerDutyIncidentWorkflowTrigger_BasicConditionalAllServices as template, if not, please wait a little until I add it to this PR.

jamesforwardnwboxed:
> Thank you for this contribution @jamesforwardnwboxed ! The main reason we did not merge this yet is because it lacks an acceptance test. Do you think you can include one? You can use TestAccPagerDutyIncidentWorkflowTrigger_BasicConditionalAllServices as template, if not, please wait a little until I add it to this PR.

Hi, apologies for late reply.

I checked and there already seems to be one where `""` is set, which is slightly confusing.

I've also adjusted my PR from `"incident.status matches 'triggered'"` to `""` and it seems to apply (although the state itself doesn't end up changing). I'm slightly confused by this as it probably requires a little more digging into the code, which unfortunately I don't have the work capacity to do.

cjgajard:
Because of the `omitempty` in the [library for API calls](https://github.com/PagerDuty/terraform-provider-pagerduty/blob/master/vendor/github.com/heimweh/go-pagerduty/pagerduty/incident_workflow_trigger.go#L19C2-L19C90) the provider was able to create triggers with an empty `condition` but it was not able to remove the value once created with a non-empty value.

I think we're ready to merge this. Can you confirm and give us a green light @jamesforwardnwboxed ?

jamesforwardnwboxed:
Hey @cjgajard sounds good to me!

Issue #1022, opened 2025-05-24
Feature request: Add `teams` attribute to data.pagerduty_escalation_policy


mdb:
Hi! Thanks for your work on `terraform-provider-pagerduty`.

I'd love for the [pagerduty_escalation_policy](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/data-sources/escalation_policy) data source to support additional computed attributes, such as `teams` (currently already supported by the [pagerduty_escalation_policy resource](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/resources/escalation_policy)).

This would unblock the ability to drive more dynamic Terraform configurations and even [custom conditions](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions) via the ability to compute the teams associated with an escalation policy.

Thanks!

PR #1023, opened 2025-05-24
feat: esc policy data source supports teams/description

mdb:
This addresses issue #1022 and adds support for `teams` and `description` computed attributes to the `data.pagerduty_escalation_policy` data source.

### DIFF
--- a/pagerdutyplugin/data_source_pagerduty_escalation_policy.go
+++ b/pagerdutyplugin/data_source_pagerduty_escalation_policy.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/PagerDuty/go-pagerduty"
 	"github.com/PagerDuty/terraform-provider-pagerduty/util/apiutil"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/path"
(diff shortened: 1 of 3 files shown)


pagerdutyplugin/data_source_pagerduty_escalation_policy.go
```diff
@@ -26,6 +27,13 @@ func (*dataSourceEscalationPolicy) Schema(ctx context.Context, req datasource.Sc
 		Attributes: map[string]schema.Attribute{
 			"id":   schema.StringAttribute{Computed: true},
 			"name": schema.StringAttribute{Required: true},
+			"description": schema.StringAttribute{
+				Computed: true,
+			},
+			"teams": schema.ListAttribute{
+				Computed:    true,
+				ElementType: types.StringType,
```
cjgajard, line 35: Thank you for your contribution! We want to keep this provider as a mere reflection of the public API whenever possible. Can you please change teams to be a [list of API objects](https://developer.pagerduty.com/api-reference/51b21014a4f5a-list-escalation-policies) instead of a list of IDs? Also, for this to work you should add `Include: []string{"teams"}` to `pagerduty.ListEscalationPoliciesOptions` at line 53.

cjgajard (changes requested):

pagerdutyplugin/data_source_pagerduty_escalation_policy.go
mdb: > Also, for this to work you should add Include: []string{"teams"} to pagerduty.ListEscalationPoliciesOptions at line 53

I don't believe I should; `teams` is compute (and not exposed as a user-facing argument). It's intended to be the complete list of all teams returned by the API. Do you see things differently?

mdb: > Also, for this to work you should add Include: []string{"teams"} to pagerduty.ListEscalationPoliciesOptions at line 53.

Done!

cjgajard: Oh, great catch! Let's keep it this way then!

cjgajard (approved):
```
+go test ./pagerdutyplugin -run TestAccDataSourcePagerDutyEscalationPolicy_ -v
=== RUN   TestAccDataSourcePagerDutyEscalationPolicy_Basic
--- PASS: TestAccDataSourcePagerDutyEscalationPolicy_Basic (18.47s)
PASS
ok      github.com/PagerDuty/terraform-provider-pagerduty/pagerdutyplugin       19.219s
```

Issue #318, opened 2021-03-25
pagerduty_team_membership cause slowness to terraform plan due to large list of team members in some teams


maxchen-000:
We have some teams with large amount of users in it. For example, we have a team called "STAKEHOLDER" that have about  1500+ users in it. When we used "pagerduty_team_membership" to create the team membership and run terraform plan, it tried to use the https://api.pagerduty.com/teams/{id}/members api to check if that user exist in that team and which role it has as part of the terraform plan. Because it has to list all users of that team to find that 1 particular user for every user, it's very slow. For example, it need to do a minimum of 15 api calls to list all 1500 users for that team using the limit of 100 (which is the PD api max limit) to list all 1500 users, then check and find if that user exist in that team. Since we have 1500 users in that team and we have 1500 "pagerduty_team_membership" terraform resources created for this. It need to make about 15*1500=22,500 minimum calls for terraform plan that we run each time. This number goes up to 90,000 if it's suing the default limit of 25. This is taking about 1 hour to run, which is very inefficient. There need to be a dedicated api to check if a user belong to a team with a particular membership so that it only need to make a single call to check team membership

### Terraform Version
Terraform v0.13.4
pagerduty v1.9.4

### Affected Resource(s)
pagerduty_team_membership

### Terraform Configuration Files
```hcl
resource "pagerduty_team_membership" "user1-STAKEHOLDER" {
   user_id = pagerduty_user.user1.id
   team_id = data.pagerduty_team.STAKEHOLDER.id
   role = "observer"
}
```

pridumre:
@maxchen-000  did you got any workaround for this issue?
we have similar issue for this https://github.com/PagerDuty/terraform-provider-pagerduty/issues/709

maxchen-000:
no, we just let it run for 1-2 hours until it's complete

PR #1108, opened 2026-03-25
Add opt-in cache for large team membership reads

yoshi-taka:
## Summary

This PR adds an opt-in in-memory cache for `pagerduty_team_membership` reads to reduce repeated `/teams/{id}/members` list calls for large teams.

This is intended to address the scaling problem described in #318, where `terraform plan` / refresh becomes very slow when a single team has a large number of memberships managed as individual `pagerduty_team_membership` resources.

The cache is enabled only when `PAGERDUTY_ENABLE_TEAM_MEMBERSHIP_READ_CACHE=1` is set.

## Background

Today, each `pagerduty_team_membership` read walks the team members list from the beginning until it finds the target user. For large teams, and especially when many resources share the same `team_id`, this causes the provider to re-read the same paginated team member list many times during a single refresh.

This PR changes the steady-state read path so that, when enabled, memberships for the same team share a single snapshot within the same Terraform run.

## What changed

- add a dedicated read-path helper for `pagerduty_team_membership`
- add an opt-in in-memory cache for team membership reads
- make cache access concurrency-safe and deduplicate in-flight snapshot fetches for the same team
- share one team snapshot across repeated steady-state reads for the same team
- invalidate the cached snapshot after `Create`, `Update`, and `Delete`
- bypass the cache for post-write verification reads
- add unit tests for cache reuse, invalidation, pagination, missing members/teams, and fetch error recovery

## Performance impact

Without this cache, if you have many membership resources for a single team, Terraform may walk the same paginated team-members list repeatedly during refresh/plan. With the cache enabled, steady-state reads for that team share a single snapshot fetch within the same provider process.

This changes the repeated-read pattern from roughly `O(resources * pages)` to `O(pages)` per team for steady-state reads, which should significantly reduce API rate-limit pressure and execution time for large teams.

## Cache scope

This is a short-lived in-memory cache scoped to a single Terraform provider process. It is only used within one Terraform run and is not persisted across separate plan/apply executions.

## Why this is opt-in

I kept this behind `PAGERDUTY_ENABLE_TEAM_MEMBERSHIP_READ_CACHE=1` instead of enabling it by default because the optimization is specifically aimed at accounts with very large teams and many `pagerduty_team_membership` resources sharing the same `team_id`.

There is an explicit tradeoff here:

- for large teams with many memberships, sharing a single team snapshot can dramatically reduce repeated list calls
- for small teams, or for configurations with only one/few membership resources, this can be slower
- the reason is that the snapshot path reads the full team member list, while the existing behavior can stop early as soon as the target user is found

Because of that tradeoff, I chose to keep the current behavior as the default and make the cache an opt-in feature for users who already know they are hitting this scaling problem.

## Scope

This PR only changes the `pagerduty_team_membership` read path. Similar list-based read patterns in other resources are out of scope for this change.

## Testing

- added unit tests covering cache reuse, invalidation, pagination, missing members/teams, and fetch error recovery
- verified the Go test suite locally
- I have not yet run a manual large-team performance comparison against a real PagerDuty account, and I do not currently have access to a suitable PagerDuty account with teams for end-to-end Terraform verification

## References

- Relates to #318

### DIFF
--- a/pagerdutyplugin/resource_pagerduty_team_membership.go
+++ b/pagerdutyplugin/resource_pagerduty_team_membership.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"net/http"
 	"net/url"
 	"strings"
 	"time"
@@ -24,7 +23,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 )
 
-type resourceTeamMembership struct{ client *pagerduty.Client }
+type resourceTeamMembership struct {
+	client *pagerduty.Client
+}
 
 var (
 	_ resource.ResourceWithConfigure   = (*resourceTeamMembership)(nil)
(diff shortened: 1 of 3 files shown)


imjaroiswebdev:
Addressed by #1111

PR #1109, opened 2026-04-06
Validates existing config before create to prevent accidental overwrites

cjgajard:
This PR adds pre-creation validation to all four \`pagerduty_event_orchestration_path_*\` resources to prevent accidental overwrites of existing orchestration configurations.

## Problem

When creating any orchestration path resource without first importing it, users could inadvertently overwrite existing PagerDuty configurations that were created outside Terraform. This leads to unexpected data loss and requires manual recovery.

## Solution

Extracted a shared \`checkExistingOrchestrationPathConfig\` function (called from both \`CustomizeDiff\` and \`resourceCreate\`) that:

1. **Validates before CREATE**: Fetches the current path configuration and blocks creation if non-trivial content already exists.
2. **Detects meaningful configuration per path type**:
   - **router**: non-trivial when catch_all routes somewhere other than \`unrouted\`, or any rules are present.
   - **global / service / unrouted**: non-trivial when catch_all has any action field set (severity, priority, annotations, automation actions, etc.) or when any rules exist. The \`suppress=true\` default that the API injects on the unrouted path is excluded from the trivial check.
3. **Provides clear guidance**: Returns a descriptive error with the exact \`terraform import\` command to resolve the conflict.
4. **Handles orphaned service resources**: Treats a 403 response (returned instead of 404 when the parent service is deleted) as "no existing config" for the service path type only.
5. **Preserves existing behavior**: Only runs during CREATE; imports and updates proceed normally.

The check retries API calls for 5 seconds to handle transient failures.

## Resources updated

- \`pagerduty_event_orchestration_service\`
- \`pagerduty_event_orchestration_global\`
- \`pagerduty_event_orchestration_router\`
- \`pagerduty_event_orchestration_unrouted\`

## Documentation

Each resource doc page gains an import-hint note explaining how to resolve the overwrite-guard error.

## Testing

Acceptance tests covering the overwrite-guard behavior were added for all four resource types (see \`TestAccPagerDutyEventOrchestrationPath*_OverwriteGuard\` in each \`*_test.go\` file). Each test:
1. Creates a fresh resource so the orchestration exists in state.
2. Removes it from Terraform state (triggering a DELETE/reset), then injects non-trivial configuration out-of-band via the API.
3. Runs \`terraform plan\` only (\`PlanOnly: true\`) and asserts the error fires from \`CustomizeDiff\` before any write reaches the API.

### DIFF
--- a/pagerduty/event_orchestration_path_util.go
+++ b/pagerduty/event_orchestration_path_util.go
@@ -448,7 +448,13 @@ func checkExistingOrchestrationPathConfig(ctx context.Context, client *pagerduty
 		path, _, err := client.EventOrchestrationPaths.GetContext(ctx, orchID, pathType)
 		if err != nil {
 			if isErrCode(err, http.StatusForbidden) {
-				return nil
+				// For service paths the API returns 403 (not 404) when the parent
+				// service has been deleted; treat it as "no existing config".
+				// For other path types a 403 is a real permission error.
+				if pathType == "service" {
+					return nil
+				}
+				return retry.NonRetryableError(err)
 			}
 			if isErrCode(err, http.StatusBadRequest) {
 				return retry.NonRetryableError(err)
@@ -483,7 +489,10 @@ func checkExistingOrchestrationPathConfig(ctx context.Context, client *pagerduty
 	} else {
 		if existingPath.CatchAll != nil && existingPath.CatchAll.Actions != nil {
 			a := existingPath.CatchAll.Actions
-			if a.Suppress || a.DropEvent || a.Priority != "" || a.Severity != "" ||
+			// The unrouted path always has suppress=true set by the API; exclude it
+			// from the trivial check so a fresh orchestration is never falsely blocked.
+			suppressNonTrivial := a.Suppress && pathType != "unrouted"
+			if suppressNonTrivial || a.DropEvent || a.Priority != "" || a.Severity != "" ||
 				a.EventAction != "" || a.Annotate != "" || a.RouteTo != "" ||
 				a.Suspend != nil || a.EscalationPolicy != nil ||
 				len(a.Variables) > 0 || len(a.Extractions) > 0 ||
--- a/pagerduty/event_orchestration_path_util.go
+++ b/pagerduty/event_orchestration_path_util.go
@@ -3,8 +3,11 @@ package pagerduty
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/heimweh/go-pagerduty/pagerduty"
 )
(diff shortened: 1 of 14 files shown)


pagerduty/event_orchestration_path_util.go
```diff
@@ -434,6 +437,88 @@ func convertEventOrchestrationPathWarningsToDiagnostics(warnings []*pagerduty.Ev
...
+			a := existingPath.CatchAll.Actions
+			// The unrouted path always has suppress=true set by the API; exclude it
+			// from the trivial check so a fresh orchestration is never falsely blocked.
+			suppressNonTrivial := a.Suppress && pathType != "unrouted"
+			if suppressNonTrivial || a.DropEvent || a.Priority != "" || a.Severity != "" ||
+				a.EventAction != "" || a.Annotate != "" || a.RouteTo != "" ||
+				a.Suspend != nil || a.EscalationPolicy != nil ||
+				len(a.Variables) > 0 || len(a.Extractions) > 0 ||
+				len(a.AutomationActions) > 0 || len(a.PagerdutyAutomationActions) > 0 ||
+				len(a.IncidentCustomFieldUpdates) > 0 {
```
metavida, line 504: **Not a Blocker:** I worry that as we enhance the Event Orchestration product in the future that it'll be easy to forget to update this portion of code when we add support for a new type of Action. I don't immediately have a good solution in mind but figured I'd mention the concern in case you had an idea.

website/docs/r/event_orchestration_global.html.markdown
```diff
@@ -173,7 +173,8 @@ The following attributes are exported:
 
 Global Orchestration can be imported using the `id` of the Event Orchestration, e.g.
 
-
 ```
 $ terraform import pagerduty_event_orchestration_global.global 1b49abe7-26db-4439-a715-c6d883acfb3e
 ```
+
+-> **Note:** If Terraform reports an error like _"has existing configuration that might be overwritten; please import this resource before creating it"_, the global orchestration path already has rules or non-default catch-all settings that Terraform does not yet manage. Run the import command above to bring that existing configuration into your Terraform state, then re-run `terraform apply`.
```
metavida, line 180: **Kudos:** I hadn't seen the `->` syntax before in Markdown/Terraform docs but did confirm that it's used elsewhere to good effect!

https://github.com/PagerDuty/terraform-provider-pagerduty/blob/95cf9d8b41aaa7f90df29938789aa8abc3fc3212/website/docs/guides/migrate-schedule-to-schedulev2.html.markdown#L27-L29

Image

via https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs/guides/migrate-schedule-to-schedulev2

pagerduty/event_orchestration_path_util.go
```diff
@@ -434,6 +437,88 @@ func convertEventOrchestrationPathWarningsToDiagnostics(warnings []*pagerduty.Ev
...
+	var existingPath *pagerduty.EventOrchestrationPath
+
+	retryErr := retry.RetryContext(ctx, 5*time.Second, func() *retry.RetryError {
+		path, _, err := client.EventOrchestrationPaths.GetContext(ctx, orchID, pathType)
+		if err != nil {
+			if isErrCode(err, http.StatusForbidden) {
+				// For service paths the API returns 403 (not 404) when the parent
+				// service has been deleted; treat it as "no existing config".
+				// For other path types a 403 is a real permission error.
+				if pathType == "service" {
```
metavida, line 454: **Kudos:** Great work spotting these type-specific behavior differences **and** documenting them well.

metavida (commented):
I don't see any hard blockers! A few notes:

1. The PR description is out of date vs the actual implementation (e.g. PR description only mentions `pagerduty_event_orchestration_path_service` but this PR correctly updates all `pagerduty_event_orchestration_path_*` resources)
2. I have **not** done any QA on this PR to confirm things actually work. My assumption is that you've done this QA yourself but I haven't directly seen any of the QA evidence so I wanted to mention it. Things I'd love to have confirmation about:
    1. How does the error actually present to users?
    2. Does the error apply both to `terraform plan` and `terraform apply`?
    3. Do updates to existing `pagerduty_event_orchestration_path_*` resources actually still work?

All that said, as I read through the code it makes sense to me!!

cjgajard (commented):
Thanks for the thorough review!

**PR description** — updated to cover all four path types, document the per-type trivial/non-trivial logic, and describe the test structure.

**Exhaustiveness of the action-field check** — added a comment in `checkExistingOrchestrationPathConfig` naming the SDK struct (`EventOrchestrationPathRuleActions`) that the check mirrors, with an explicit note that `DynamicRouteTo` is intentionally absent (router-only; the API never sets it on non-router catch_all). That comment should serve as the "update me" signal when new action fields land in go-pagerduty.

**QA evidence** — addressing your three questions directly:

1. **How does the error present?** It surfaces as a Terraform plan-time error (from `CustomizeDiff`) with the message: `"the <type> orchestration (ID: <id>) has existing configuration that might be overwritten; please import this resource before creating it using: terraform import <resource_type>.<resource_name> <id>"`. The `PlanOnly: true` step in each `_OverwriteGuard` acceptance test exercises this exact path and asserts the error text.

2. **Does it apply to both `terraform plan` and `terraform apply`?** Yes. The check runs in `CustomizeDiff`, which Terraform invokes during both plan and apply. It also runs at the top of each `resourceCreate` function as a second line of defence (catches the apply-without-prior-plan edge case). The acceptance test uses `PlanOnly: true` to confirm the plan-time firing; apply would be blocked by the same `CustomizeDiff` call.

3. **Do updates to existing resources still work?** Yes. All four `customizeDiff` functions guard with `if diff.Id() != "" { return nil }`, so the check is skipped entirely for any resource that already exists in state. The pre-existing `Basic` acceptance tests for each resource type exercise multi-step in-place updates and confirm no regression.

imjaroiswebdev (approved):

PR #1112, opened 2026-04-15
Add SizeAtMost(1) validator to use_app_oauth_scoped_token block

guineveresaenger:
According to https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks, when migrating the schema config block to Plugin Framework, a size validator is needed to reflect the equivalent of `MaxItems: 1` in the [corresponding SDKv2 block](https://github.com/akamai/terraform-provider-akamai/blob/9dbd942f4a48cfdb27448fe7809305ece6783453/pkg/akamai/sdk_provider.go#L38).

We discovered this downstream when consuming the Plugin Framework schema, and our model inaccurately did not flatten the config block.

Thank you for considering this pull request!

Refs:
- pulumi/pulumi-pagerduty#569 (downstream bug report)

### DIFF
--- a/pagerdutyplugin/provider.go
+++ b/pagerdutyplugin/provider.go
@@ -8,12 +8,14 @@ import (
 	"strings"
 
 	"github.com/PagerDuty/go-pagerduty"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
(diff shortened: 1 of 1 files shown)


cjgajard (approved):
LGTM!

PR #1134, opened 2026-07-06
Fix DMS error detection to match real API wording (FDE-894)

pdt-rmunoz:
## Summary
- Live testing against staging surfaced that the API returns "...triage service" wording for `pagerduty_service_integration` create and `pagerduty_maintenance_window` create/update, not the `"Account Default Mobilization Service"` substring `IsDefaultMobilizationServiceError` checked for.
- As a result, the guardrails added in this branch silently never fired for those two operations, and users still saw the raw API error instead of the friendly message.
- Broadened detection to check for both known wordings.

## Test plan
- [x] `go test ./util/...` passes, including two new cases reproducing the exact staging 422 responses observed
- [x] Verified locally against `api.pd-staging.com` with a dev_overrides build:
  - `pagerduty_service_integration` create on the DMS now returns the friendly `DMSMsgServiceIntegrationCreate` message
  - `pagerduty_maintenance_window` create on the DMS now returns the friendly `DMSMsgMaintenanceWindow` message

### DIFF
--- a/util/default_mobilization_service.go
+++ b/util/default_mobilization_service.go
@@ -5,13 +5,18 @@ import (
 	"strings"
 )
 
-// DefaultMobilizationServiceErrSignal is the substring the PagerDuty API includes
-// in its error whenever an operation is rejected for targeting the account's
-// Default Mobilization Service (formerly the "Triage service"). Both the heimweh
-// and PagerDuty go-pagerduty clients embed the API message in their Error()
-// output, so this single substring check identifies the condition regardless of
-// which client produced the error, without an extra lookup call.
-const DefaultMobilizationServiceErrSignal = "Account Default Mobilization Service"
+// defaultMobilizationServiceErrSignals are the substrings the PagerDuty API is
+// known to include in its error whenever an operation is rejected for targeting
+// the account's Default Mobilization Service. Both the heimweh and PagerDuty
+// go-pagerduty clients embed the API message in their Error() output, so a
+// substring check identifies the condition regardless of which client produced
+// the error, without an extra lookup call. The API uses inconsistent wording
+// across endpoints (e.g. "triage service" for integration and maintenance
+// window creation), so we check for all known variants.
+var defaultMobilizationServiceErrSignals = []string{
+	"Account Default Mobilization Service",
+	"triage service",
+}
 
 // defaultMobilizationServiceSuffix is the shared trailing guidance appended to
 // every Default Mobilization Service message so operators get consistent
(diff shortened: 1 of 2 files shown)


imjaroiswebdev (changes requested):
Thanks for catching this — the `"triage service"` wording gap is a real bug, and the repro tests against the staging 422s are exactly what I'd want to see. The core fix is right; one change requested before merge, plus two non-blocking nits.

### 1. (Change requested) Matching the bare substring `"triage service"` is over-broad, and this function gates retry decisions

`IsDefaultMobilizationServiceError` isn't only used to swap the message — it also gates `retry.NonRetryableError(...)` in both `event_orchestration_path_*` update paths, where **400 is non-retryable and everything else (including a 422) is retryable**. So a false positive there doesn't just show the wrong text; it converts a genuinely retryable error into a hard failure, so an apply that might have succeeded on retry now fails — with a misleading "Default Mobilization Service" message.

`"triage service"` is generic enough to appear in unrelated error text: PagerDuty service names are user-chosen, so a service named e.g. `payments triage service` echoed back in some other validation failure would match. The previous `"Account Default Mobilization Service"` signal didn't have that surface.

Suggest anchoring to the fuller confirmed phrases rather than the bare words, e.g.:

```go
var defaultMobilizationServiceErrSignals = []string{
	"Account Default Mobilization Service",
	"cannot be created on a triage service",   // service_integration create
	"cannot include the triage service",       // maintenance_window create/update
}
```

If you'd rather keep it broad to catch wording not yet observed, that's a reasonable call too — but then let's add a guard so the retry short-circuit only trips on the errors we actually mean.

### 2. (Non-blocking) Case sensitivity

`strings.Contains` is case-sensitive and the signals are lower-case. If any endpoint returns `"Triage service"` / `"Triage Service"`, detection silently misses again — the same failure mode this PR fixes. Consider normalizing: `strings.Contains(strings.ToLower(msg), signal)` with lower-cased signals.

### 3. (Non-blocking) Add a negative test

The two new cases only assert `want: true`. A negative case — an unrelated error that mentions a differently-named "triage service" and should **not** match — would lock in the precision of whichever matcher you land on, so a future over-broadening gets caught.

pdt-rmunoz:
Thanks for the thorough review — all three addressed in c94e1740:

**1. (Change requested) Anchored to full phrases** — good catch on the retry-gating risk. Replaced the bare `"triage service"` with the two confirmed full phrases:

```go
var defaultMobilizationServiceErrSignals = []string{
    "account default mobilization service",
    "cannot be created on a triage service",
    "cannot include the triage service",
}
```

**2. (Non-blocking) Case sensitivity** — now lower-cases both the error message and the signals before comparing, so `"Triage Service"` / `"TRIAGE SERVICE"` etc. still match.

**3. (Non-blocking) Negative test** — added a case with an unrelated 422 mentioning a user-named `payments triage service` service, asserting `want: false`, plus a casing test asserting the different-case variant still matches.

`go build ./...`, `go vet`, and `go test ./util/...` all pass.

pdt-rmunoz (commented):
please check again

imjaroiswebdev:
LGTM! Thanks!

imjaroiswebdev (approved):

PR #1136, opened 2026-07-07
Add `id` lookup to `pagerduty_service` data source

### LINKED ISSUE #1135: data/pagerduty_service: support lookup by `id` (not only `name`)
### Use case

The `pagerduty_service` data source currently only supports lookup by **name** (`name` is `Required`; `id` is `Computed`/output-only). In my setup services are managed declaratively and their **IDs are already pinned in my YAML/config**, so I reliably have the service ID on hand but not always a guaranteed-unique name to search by. Today I'm forced to look a service up by name even when I already know its ID.

### Expected Behavior

Be able to resolve a service by `id`, with the existing name-based lookup still supported:

```hcl
data "pagerduty_service" "by_id" {
  id = "PXXXXXX"
}

data "pagerduty_service" "by_name" {
  name = "My Service"
}
```

### Actual Behavior

`id` is output-only; only `name` (Required) can be supplied. There is no way to fetch a service by ID through the data source — you must search by name, which is a paginated `ListServices` + client-side filter.

### Proposed solution

Make `id` and `name` both `Optional` + `Computed` with an "exactly one of" constraint, and in `Read`, when `id` is provided, call the client's direct `GetServiceWithContext(ctx, id, …)` instead of the paginated `ListServices` search.

This mirrors a pattern **already in this provider**: the `pagerduty_license` data source (`pagerdutyplugin/data_source_pagerduty_license.go`) already accepts either `id` or `name` (both `Optional`+`Computed`). The vendored `go-pagerduty` client already exposes `GetServiceWithContext` (used by the service *resource*), so no new API plumbing is required. As a bonus, fetching by ID is a single API call rather than paginating every service — cheaper on large accounts.

### Willing to contribute

Happy to open a PR (schema + `Read` branch + acceptance test + docs) following the `pagerduty_license` precedent. I wanted to confirm the design — **id-or-name, mutually exclusive** — is what you'd want before implementing. cc @cjgajard, who migrated this data source to the plugin framework, in case there's prior context.

### Environment

- Provider: `PagerDuty/pagerduty` v3.33.x / current `master`
- Affected: data source `pagerduty_service`

### References
- In-repo precedent: `pagerdutyplugin/data_source_pagerduty_license.go` (id-or-name lookup)
- Client method: `go-pagerduty` `GetServiceWithContext`

nikolauspschuetz:
Closes #1135.

### What

Adds the ability to look up a service by **`id`** in the `pagerduty_service` data source. Previously `name` was required and `id` was output-only, so you had to search by name even when the service ID was already known.

### How

- Schema: `id` and `name` are now both `Optional` + `Computed`; exactly one must be provided.
- `Read`: when `id` is set, the service is fetched with a single direct `GetServiceWithContext` request; otherwise the existing paginated name search runs unchanged.
- This mirrors the id-or-name pattern already used by the `pagerduty_license` data source (`pagerdutyplugin/data_source_pagerduty_license.go`), and `GetServiceWithContext` is already vendored and used by the service resource — no new API plumbing or dependencies.

### Why

Services are commonly managed declaratively with their IDs pinned in config, so lookup-by-ID is a natural need. It's also cheaper — one direct GET vs. paginating every service in the account.

### Backward compatibility

Fully backward compatible: existing `name`-based configs are unchanged (`name` simply moved from Required to Optional; `id` gains an input role in addition to being computed).

### Tests

- Added `TestAccDataSourcePagerDutyService_ByID`, which creates a service and resolves it via `id`, asserting the data source returns the same attributes as the resource.
- `go build ./...`, `go vet`, and `gofmt` are clean; the new acceptance test compiles and is skipped without `TF_ACC`.
- I don't have a PagerDuty test token on hand right now, so I haven't run the acceptance test live — happy for CI (or a maintainer) to run it, and I'll follow up if anything needs adjusting.

<sub>Developed with AI assistance (Claude Code); I directed the change and verified build/vet/fmt + test compilation locally.</sub>


nikolauspschuetz:
Hi @cjgajard @imjaroiswebdev — gentle nudge on this one. It adds direct `id` lookup to the `pagerduty_service` data source (a single `GetService` call instead of paginating + name-matching), following the same id-or-name shape as the `pagerduty_license` data source you authored, @cjgajard. Green + has an acceptance test; happy to adjust anything. Thanks!

PR #1140, opened 2026-07-28
[DO NOT MERGE UNTIL GA] UA-1146: Add ua_runner to runner_type validation enum

ezhuango:
- Adds `ua_runner` to the `runner_type` validation enum in
  `resource_pagerduty_automation_actions_runner.go`
  - This allows Terraform to recognize `ua_runner` as a valid runner type for
  import/read operations
  - Creation via Terraform is still restricted to `runbook` type only (ua_runners are
   created through the Connectors UI)


cjgajard (approved):
LGTM!🎉 Thank you @ezhuango!

PR #1141, opened 2026-08-05
[TFPROVDEV-146] Handle out-of-band deletion of pagerduty_user_contact_method

imjaroiswebdev:
**Jira:** [TFPROVDEV-146](https://pagerduty.atlassian.net/browse/TFPROVDEV-146)

## Change intentions (2 of max 2)

1. Honor the caller-supplied `errCallback` in `fetchPagerDutyUserContactMethod`, which was accepted as a parameter and then ignored.
2. Return a clear, actionable error when an update targets a contact method that no longer exists, instead of a bare `404 Not Found`.

## Summary

Reported via support: after deleting a Terraform-managed contact method in the PagerDuty web interface, an operator hit a `404 Not Found` that halted the run, and worked around it with `terraform state rm pagerduty_user_contact_method...` on every occurrence. The report asked for the read path to return "null" on 404 so Terraform could self-heal.

**The read path already does that**, on `master` and on the reported 3.31.1 alike — `Read` → `fetchPagerDutyUserContactMethod` → `handleNotFoundError` clears the ID on both the structured 404 and the malformed-string 404, so a refresh-enabled `plan`/`apply` re-creates the contact method with no manual state surgery. There was no missing exception handler there, and this PR does not change that behavior — it adds an acceptance test that pins it down (see Testing), because the resource had **no** acceptance test coverage at all before this.

Two real gaps did turn up in the same resource while confirming that:

**1. `errCallback` was ignored (`resource_pagerduty_user_contact_method.go`)**

```go
// before
errResp := handleNotFoundError(err, d)
// after
errResp := errCallback(err, d)
```

Every other `fetch*` helper in the package (`fetchService`, `fetchSchedule`, `fetchEscalationPolicy`, `fetchPagerDutyUserNotificationRule`, …) calls the callback; this was the lone outlier. The visible consequence is on create: `Create` passes `genError` precisely so that a 404 on the read-back is retried and then reported, but the hardcoded `handleNotFoundError` instead treated it as "resource gone" and silently dropped the just-created contact method from state.

**2. `Update` surfaced the raw API error**

Terraform only plans an update against an ID it read during refresh, so a 404 here means the object was deleted **after** the plan was generated — i.e. the run applied a saved plan file, or used `-refresh=false`. That is the one path that produces the reported symptom, and the remedy is a refresh, not `terraform state rm`. New `handleStaleIDError` (in `provider.go`, next to `handleNotFoundError`) says so:

```
pagerduty_user_contact_method PC456 no longer exists in PagerDuty, so the planned
update could not be applied. It was most likely deleted outside of Terraform after
this plan was generated, which happens when applying a saved plan file or running
with `-refresh=false`. Run `terraform apply -refresh-only`, or a plain
`terraform plan`/`terraform apply` with refresh enabled, and Terraform will forget
the deleted object and re-create it from your configuration. Removing it from state
with `terraform state rm` is not required.

Original API error: PUT API call to ... failed 404 Not Found. Code: 2100, ...
```

Non-404 errors pass through untouched, so a 500 or a permissions failure is never misreported as an out-of-band deletion. The original API error is preserved for debuggability, matching the `util.DefaultMobilizationServiceMsg` pattern from #1133.

Not touched, for scope: the `Import` path still returns the raw client error for a bad ID. The report also mentioned "similar concerns for Schedules and other resources" — `fetchSchedule` and the rest already call `errCallback` correctly, so there is nothing analogous to fix there.

## Docs preview

The new note in `website/docs/r/user_contact_method.html.markdown` renders directly below the existing "behaves a little differently" callout on the registry page:

> [!NOTE]
> If a contact method managed by Terraform is deleted directly in the PagerDuty web interface, a normal `terraform plan`/`terraform apply` detects that it is gone during refresh and re-creates it. Removing it from state with `terraform state rm` is not needed. If your workflow applies a saved plan file or runs with `-refresh=false`, the deletion cannot be detected before the apply, so run `terraform apply -refresh-only` (or re-plan with refresh enabled) first.

## Testing

- `go build ./...`, `go vet ./pagerduty/`, `gofmt -l pagerduty/` — all clean

- Unit tests — `handleStaleIDError` against a structured `*pagerduty.Error` 404, the malformed-string 404 the client falls back to when the body does not decode, a nil error, and a non-404 passthrough — **4/4 pass**:

  ```
  $ go test ./pagerduty/ -run 'TestHandleStaleIDError' -v
  === RUN   TestHandleStaleIDError
  === RUN   TestHandleStaleIDError/structured_error
  === RUN   TestHandleStaleIDError/malformed_error
  === RUN   TestHandleStaleIDError/nil_error
  === RUN   TestHandleStaleIDError/other_error_passes_through
  --- PASS: TestHandleStaleIDError (0.00s)
      --- PASS: TestHandleStaleIDError/structured_error (0.00s)
      --- PASS: TestHandleStaleIDError/malformed_error (0.00s)
      --- PASS: TestHandleStaleIDError/nil_error (0.00s)
      --- PASS: TestHandleStaleIDError/other_error_passes_through (0.00s)
  PASS
  ok  	github.com/PagerDuty/terraform-provider-pagerduty/pagerduty	1.037s
  ```

- Acceptance tests — new file, run against a live account. `_RecreateAfterOutOfBandDelete` reproduces the report by deleting the contact method through the API between steps (standing in for the web interface), re-applying the same config, and asserting a **new** ID lands in state, proving re-creation rather than a stale read:

  ```
  $ TF_ACC=1 go test ./pagerduty/ -run 'TestAccPagerDutyUserContactMethod_' -v -timeout 30m
  === RUN   TestAccPagerDutyUserContactMethod_Basic
  --- PASS: TestAccPagerDutyUserContactMethod_Basic (21.42s)
  === RUN   TestAccPagerDutyUserContactMethod_RecreateAfterOutOfBandDelete
  --- PASS: TestAccPagerDutyUserContactMethod_RecreateAfterOutOfBandDelete (20.73s)
  PASS
  ok  	github.com/PagerDuty/terraform-provider-pagerduty/pagerduty	42.731s
  ```

`_Basic` also covers the update path this PR modifies (address + label change).

[TFPROVDEV-146]: https://pagerduty.atlassian.net/browse/TFPROVDEV-146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-## v3.34.1 (Unreleased)
+## Unreleased
+
+BREAKING CHANGES
+* `resource/pagerduty_user_contact_method`: `user_id` is now `ForceNew`. A contact method's ID is scoped under its user, so changing `user_id` previously planned an in-place update that always failed with `404 Not Found`; it now replaces the resource ([1141](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/1141))
 
 BUG FIXES
 * `resource/pagerduty_user_contact_method`: Honor the caller's error handler when reading back a contact method, so a `404` right after create is retried and then reported instead of being silently treated as "resource gone" and dropped from state ([1141](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/1141))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## v3.34.1 (Unreleased)
+
+BUG FIXES
+* `resource/pagerduty_user_contact_method`: Honor the caller's error handler when reading back a contact method, so a `404` right after create is retried and then reported instead of being silently treated as "resource gone" and dropped from state ([1141](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/1141))
+
+ENHANCEMENTS
+* `resource/pagerduty_user_contact_method`: Return a clear, actionable error when an update targets a contact method that was deleted outside of Terraform after the plan was generated, pointing at `terraform apply -refresh-only` instead of surfacing a bare `404 Not Found` ([1141](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/1141))
+
+DOCS
+* `resource/pagerduty_user_contact_method`: Document that a contact method deleted in the web interface is re-created by a refresh-enabled plan/apply and does not require `terraform state rm` ([1141](https://github.com/PagerDuty/terraform-provider-pagerduty/pull/1141))
+
 ## v3.34.0 (Jul 7, 2026)
 
 ENHANCEMENTS
--- a/pagerduty/import_pagerduty_user_contact_method_test.go
+++ b/pagerduty/import_pagerduty_user_contact_method_test.go
@@ -1,0 +1,38 @@
+package pagerduty
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+func TestAccPagerDutyUserContactMethod_import(t *testing.T) {
+	username := fmt.Sprintf("tf-%s", acctest.RandString(5))
+	email := fmt.Sprintf("%s@foo.test", username)
+	address := fmt.Sprintf("%s@foo.test", acctest.RandString(6))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckPagerDutyUserContactMethodDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckPagerDutyUserContactMethodConfig(username, email, address, "Work"),
+			},
+			{
+				ResourceName:      "pagerduty_user_contact_method.foo",
+				ImportStateIdFunc: testAccCheckPagerDutyUserContactMethodID,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func testAccCheckPagerDutyUserContactMethodID(s *terraform.State) (string, error) {
+	cm := s.RootModule().Resources["pagerduty_user_contact_method.foo"].Primary
+	return fmt.Sprintf("%v:%v", cm.Attributes["user_id"], cm.ID), nil
+}
--- a/pagerduty/provider.go
+++ b/pagerduty/provider.go
@@ -222,15 +222,21 @@ func genError(err error, d *schema.ResourceData) error {
 	return fmt.Errorf("Error reading: %s: %s", d.Id(), err)
 }
 
-// handleStaleIDError converts a not-found error returned by a write call into
-// actionable guidance. Terraform only plans an update against an ID it read
-// during refresh, so a 404 on update means the object was deleted outside of
+// handleStaleIDErrorOnUpdate converts a not-found error returned by an update
+// call into actionable guidance. Terraform only plans an update against an ID it
+// read during refresh, so a 404 here means the object was deleted outside of
 // Terraform *after* the plan was generated — typically by applying a saved plan
 // file or running with refresh disabled. A refresh is all that's needed for
 // Terraform to drop the stale ID and re-create the object, so say so instead of
 // surfacing a bare API error that reads like a provider failure and pushes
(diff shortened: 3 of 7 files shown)


cjgajard (commented):
## Review

Verified against `master`: the read path already self-heals on 404 via `handleNotFoundError` (`provider.go`), and the `fetchPagerDutyUserContactMethod` fix brings it in line with every other `fetch*` helper (`fetchService`, `fetchPagerDutyUserNotificationRule`) — no issues there. One correctness gap on the new `Update` guidance, plus a few optional nits.

### Required

**1. A `user_id` change routes through `Update` and gets actively wrong guidance** — `pagerduty/resource_pagerduty_user_contact_method.go:28-31`, `:217`

`user_id` is `Required` with no `ForceNew`, and `customizeDiffResourceUserContactMethod` doesn't force replacement either, so editing `user_id` plans an in-place update: `client.Users.UpdateContactMethod(newUserID, d.Id(), ...)` -> `PUT /users/<new>/contact_methods/<old id>`. That 404s, and the new wrapper reports it as "deleted outside of Terraform after this plan was generated ... run `terraform apply -refresh-only`". That advice can't converge here: `Read` looks the method up under `d.Get("user_id")` from state (line 150), finds it alive under the *old* user on refresh, and the same update gets replanned — same error, every time. Before this PR the operator got an unhelpful bare 404; now they get a confident, wrong diagnosis pointing them at a refresh that will never fix it.

Suggested fix (a contact method can't move between users, matching `resource_pagerduty_team_membership.go:28-30`):

```go
"user_id": {
    Type:     schema.TypeString,
    Required: true,
    ForceNew: true,
},
```

Minimal alternative if you'd rather not touch the schema:

```go
if _, _, err := client.Users.UpdateContactMethod(userID, d.Id(), contactMethod); err != nil {
    if d.HasChange("user_id") {
        return err
    }
    return handleStaleIDError(err, d, "pagerduty_user_contact_method")
}
```

### Optional

**2. No test reaches the new `Update` branch** — `pagerduty/provider_test.go:46`, `pagerduty/resource_pagerduty_user_contact_method_test.go`

`TestHandleStaleIDError` calls the helper directly, and `_Basic` only exercises a *successful* update, so the call site itself (right `resourceType` string, right placement) is uncovered. A post-plan 404 is awkward to stage in the acceptance framework, but if #1 lands as `ForceNew`, a step that changes `user_id` and asserts replacement pins that path shut cheaply. Also, since this is the resource's first test file: worth adding `import_pagerduty_user_contact_method_test.go` with an `ImportStateIdFunc` returning `<user_id>:<id>`, matching `import_pagerduty_business_service_subscriber_test.go:24-29` — `Import` is explicitly scoped out of this PR and is now the only entry point with zero coverage.

**3. `handleStaleIDError`'s name is generic, its message isn't** — `pagerduty/provider.go:233`

It sits next to `handleNotFoundError` as a package-wide helper, but hardcodes "the planned update could not be applied" and the `-refresh-only` remedy. If a future caller wires it into `Create` or `Delete`, the emitted text won't describe what actually happened. Consider scoping the name (e.g. `handleStaleIDErrorOnUpdate`) or taking the operation as a parameter.

**4. CHANGELOG heading pre-assigns the version** — `CHANGELOG.md:1`

The repo convention (see `edd514b2`) is to land `## Unreleased` in the PR and promote it to a version number in a separate cut commit. `## v3.34.1 (Unreleased)` jumps that — and with an ENHANCEMENTS entry alongside the BUG FIXES one, the eventual cut may end up `3.35.0` rather than a patch bump.

imjaroiswebdev:
All four addressed in `22ab7479`.

**1.** Confirmed, and reproduced: `PUT /users/<new>/contact_methods/<old id>` 404s, and the wrapper blamed out-of-band deletion for something a refresh can't fix. `user_id` is now `ForceNew` — took that over the `d.HasChange` variant, since a contact method can't move users. Called out as a breaking change in the CHANGELOG.

**2.** Added `_UserIDForceNew` (changes `user_id`, asserts replacement; fails without the `ForceNew` above) and `import_pagerduty_user_contact_method_test.go` with an `ImportStateIdFunc` returning `<user_id>:<id>`.

**3.** Renamed `handleStaleIDErrorOnUpdate`. Kept the message fixed rather than parameterizing the operation — one call site. Doc comment now states the `ForceNew` precondition.

**4.** Back to `## Unreleased`.

Acceptance tests pass: `_import`, `_Basic`, `_RecreateAfterOutOfBandDelete`, `_UserIDForceNew`.

cjgajard (approved):
LGTM! 🎉 thank you @imjaroiswebdev

PR #1143, opened 2026-08-14
Anchor start_time_of_day validation so junk-wrapped times are rejected

nikolauspschuetz:
## What

`pagerduty_schedule` validates each layer restriction's `start_time_of_day` (HH:MM:SS) with an **unanchored** regex passed to `validation.StringMatch`:

```go
ValidateFunc: validation.StringMatch(regexp.MustCompile(`([0-1][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]`), "must be of 00:00:00 format"),
```

`validation.StringMatch` does a **substring** match, so the missing `^...$` anchors let junk-wrapped values validate as OK. All of these currently pass:

- `"12:30:00xyz"`
- `"xyz08:00:00"`
- `"99:08:00:00"`

## Fix

Extract the check into a named, **anchored** `util.ValidateTimeOfDay` validator (mirroring the existing `util.ValidateRFC3339`), and wire the schedule resource to it. The regex becomes `^([0-1][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$`.

## Test (red → green)

Added `TestValidateTimeOfDay` in `util/util_test.go`. It exercises the production validator directly.

- **Red** (regex left unanchored): `12:30:00xyz`, `xyz08:00:00`, `99:08:00:00` are accepted → test FAILS.
- **Green** (anchored): valid times (`08:00:00`, `00:00:00`, `23:59:59`) pass; junk-wrapped and out-of-range (`24:00:00`, `12:60:00`) are rejected → test PASSES.

```
$ go test ./util/ -run TestValidateTimeOfDay
ok  github.com/PagerDuty/terraform-provider-pagerduty/util
```

Pure unit test — no acceptance test or API credentials required.

## Notes

- No CHANGELOG entry added — the changelog looks maintainer-cut per release with PR numbers; happy to add an entry in your preferred section/format.
- for the record: patch drafted with AI assistance, reviewed and verified by me.

### DIFF
--- a/pagerduty/resource_pagerduty_schedule.go
+++ b/pagerduty/resource_pagerduty_schedule.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"regexp"
 	"strings"
 	"time"
 
@@ -154,7 +153,7 @@ func resourcePagerDutySchedule() *schema.Resource {
 									"start_time_of_day": {
 										Type:         schema.TypeString,
 										Required:     true,
-										ValidateFunc: validation.StringMatch(regexp.MustCompile(`([0-1][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]`), "must be of 00:00:00 format"),
+										ValidateFunc: util.ValidateTimeOfDay,
 									},
 
 									"start_day_of_week": {
--- a/util/util.go
+++ b/util/util.go
@@ -58,6 +58,23 @@ func GenErrorTimeFormatRFC339(value, k string) error {
 	return fmt.Errorf("%s is not a valid format for argument: %s. Expected format: %s (RFC3339)", value, k, time.RFC3339)
 }
 
+// timeOfDayRegex matches a 24-hour HH:MM:SS time of day. The pattern is
+// anchored so that only the whole string is a valid time; without the anchors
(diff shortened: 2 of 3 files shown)


cjgajard:
Values "99:08:00:00" and "08:00:00:99" are rejected by the API, but it does with a `500 Internal Server Error` preventing an early exit and leaving the provider retry with invalid values. This PR helps receiving an early error message.

Meanwhile the API cleans the values "12:30:00xyz" and "xyz08:00:00" before saving. This PR
might be a breaking change for some customers 😢.

nikolauspschuetz:
Good catch on the 500 — early validation is the win there.

On the cleaned values though: \`start_time_of_day\` is flattened straight back from the API (`resource_pagerduty_schedule.go:593`) and has no `DiffSuppressFunc` — unlike `start` / `rotation_virtual_start`. So a config with `"xyz08:00:00"` stores `"08:00:00"` and then re-plans a diff on every run. Those users are already in a perpetual-diff state rather than a working one, so anchoring turns a silent drift into an explicit error at plan time.

Happy to narrow it either way if you would rather not change that behaviour in a patch release — e.g. keep the anchors but tolerate surrounding whitespace, or land it behind the next major. Your call.

PR #1144, opened 2026-08-24
Read webhook_subscription secret response

cjgajard:
### What

Expose the webhook subscription shared secret as a new computed, sensitive attribute
`delivery_method.secret`, so practitioners can wire the `X-PagerDuty-Signature` verification
secret into downstream config instead of copying it out of the PagerDuty UI by hand.

### Why it is not a plain passthrough

The REST API returns `delivery_method.secret` **only** in the response to the create request.
Every subsequent `GET` and `PUT` returns it empty. Two consequences drive the implementation:

- `Create` now persists `delivery_method` from the create response before the follow-up
  `Read` runs, otherwise the secret is lost immediately.
- `setWebhookResourceData` keeps the stored secret when the API returns an empty one, so
  refreshes and updates do not clear it or produce a perpetual diff.

### Limitations

Subscriptions brought in with `terraform import` have an empty `secret` -- the value is
unrecoverable through the API. The import acceptance test therefore ignores that attribute
via `ImportStateVerifyIgnore`.

### Tests

- `TestAccPagerDutyWebhookSubscription_Basic` asserts the secret is set after create and
  still set after an in-place update.
- `TestAccPagerDutyWebhookSubscription_import` ignores `delivery_method.0.secret`.

### Docs

`website/docs/r/webhook_subscription.html.markdown` documents the attribute and its create-only
caveat, adds a sensitive `output` example, and fixes the Attributes Reference section, which
was copy-pasted from the Slack connection resource (`source_name`, `channel_name`).


website/docs/r/webhook_subscription.html.markdown
```diff
@@ -96,9 +102,8 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-  * `id` - The ID of the slack connection.
```
imjaroiswebdev, line 99: **Kudos**: Good catch! TY!

pagerduty/import_pagerduty_webhook_subscription_test.go
```diff
@@ -28,6 +28,10 @@ func TestAccPagerDutyWebhookSubscription_import(t *testing.T) {
 				ResourceName:      "pagerduty_webhook_subscription.foo",
 				ImportState:       true,
 				ImportStateVerify: true,
+				// The API returns delivery_method.secret only in the response to the
+				// creation request, so it cannot be recovered by importing an existing
+				// webhook subscription.
+				ImportStateVerifyIgnore: []string{"delivery_method.0.secret"},
```
imjaroiswebdev, line 34: TY!

imjaroiswebdev (approved):
Good addition! 🚀

PR #1146, opened 2026-08-25
TFPROVDEV-150: Add `incident_type` trigger support to `pagerduty_incident_workflow_trigger`

cjgajard:
## Summary

`pagerduty_incident_workflow_trigger` only accepted `manual` and `conditional` for its `type`
attribute, but the PagerDuty REST API (`POST /incident_workflows/triggers`) also supports
`incident_type`, which starts a workflow whenever an incident of a selected type is created or
set. This closes that gap across the SDKv2 resource, its vendored go-pagerduty client, and its
tests/docs, per PagerDuty's public OpenAPI spec (`trigger_type` enum `["conditional", "manual",
"incident_type"]`, plus a flat `incident_types` string array on the trigger object).

## Changes

- `vendor/github.com/heimweh/go-pagerduty/pagerduty/incident_workflow_trigger_type.go`: add
  `IncidentWorkflowTriggerTypeIncidentType` as a new trailing `iota` value (preserves existing
  ordinals) and wire it into both the to-string and from-string maps.
- `vendor/github.com/heimweh/go-pagerduty/pagerduty/incident_workflow_trigger.go`: add
  `IncidentTypes []string` (`json:"incident_types,omitempty"`) to `IncidentWorkflowTrigger`.
- `pagerduty/resource_pagerduty_incident_workflow_trigger.go`:
  - Add `"incident_type"` to the `type` attribute's allowed values.
  - Add a new optional `incident_types` attribute (`TypeList` of `TypeString`), not `ForceNew`.
  - Populate `IncidentTypes` in `buildIncidentWorkflowTriggerStruct` unconditionally (outside the
    `forUpdate` branch) so list changes are sent on update, not just on create.
  - Broaden the `condition` guard to reject `condition` for any type other than `conditional`
    (previously only rejected it for `manual`), matching the spec.
  - Flatten `IncidentTypes` into state in `flattenIncidentWorkflowTrigger`, only when non-empty,
    so pre-existing `manual`/`conditional` triggers get no spurious diff.
- `pagerduty/resource_pagerduty_incident_workflow_trigger_test.go`: update
  `TestAccPagerDutyIncidentWorkflowTrigger_BadType`'s expected error message for the new
  three-value enum; add `TestAccPagerDutyIncidentWorkflowTrigger_BasicIncidentType` (create +
  update of `incident_types`) and `TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithIncidentType`.
- `pagerduty/import_pagerduty_incident_workflow_trigger_test.go`: add
  `TestAccPagerDutyIncidentWorkflowTrigger_import_incidentType` covering `ImportStateVerify: true`
  for an `incident_type` trigger.
- `website/docs/r/incident_workflow_trigger.html.markdown`: document the new `type` value and the
  `incident_types` attribute, and add an `incident_type` example.
- `CHANGELOG.md`: add a `v3.36.0 (Unreleased)` `FEATURES` entry for this resource.

## Test plan

- [x] `make build && make vet && make fmtcheck && make test` pass.
- [x] `go build ./...` compiles against the patched vendor tree.
- [x] `go test ./pagerduty/... -run TestAccPagerDutyIncidentWorkflowTrigger -v` (TF_ACC unset)
      passes with all cases, including the two `incident_type` and import tests, correctly
      skipping.
- [x] Live `testacc` run (requires `TF_ACC=1` and `PAGERDUTY_TOKEN`, hits real infrastructure —
      run only with explicit credentials): `TF_ACC=1 go test ./pagerduty -run
      'TestAccPagerDutyIncidentWorkflowTrigger' -v -timeout 30m`.
- [ ] Manual smoke: `terraform apply` a two-trigger config (one `conditional`, one
      `incident_type`) and confirm the second appears in the PagerDuty UI bound to the chosen
      incident type.

## Acceptance criteria

- [x] `terraform apply` on a config with `type = "incident_type"` and a non-empty
      `incident_types` list creates the trigger, sending `trigger_type: "incident_type"` and
      `incident_types: ["<id>"]`.
- [x] Changing the `incident_types` list on an existing incident-type trigger produces a
      non-empty plan and the updated list is sent on update.
- [x] `terraform plan` with `type = "incident_type"` and a `condition` fails at plan time.
- [x] `terraform plan` with `type = "incident_type"` and `permissions { restricted = true }`
      fails with the existing "restricted can only be true when trigger type is manual" error.
- [x] A `manual`/`conditional` trigger created before this change produces an empty plan after
      upgrade.
- [x] `terraform import` of an incident-type trigger round-trips with `ImportStateVerify: true`.
- [x] `make test`, `make vet`, and `make fmtcheck` pass; the build compiles against the patched
      vendor tree.

## Vendor patch fragility (please read before merging)

`vendor/github.com/heimweh/go-pagerduty` was hand-patched to add the `incident_type` enum member
and the `incident_types` field because upstream `heimweh/go-pagerduty` has neither, and `go.mod`
is already pinned at upstream's master HEAD (`v0.0.0-20250801140645-0b96cfc9bf17`; the repo has
zero tags). `go.mod`/`go.sum` are intentionally untouched by this PR.

Consequence: any future `go mod vendor` or `make update-go-pagerduty` run will silently revert
both edits, and the build will keep compiling because `-mod=vendor` does not hash-verify vendor
contents against upstream — there will be no error, just a regression. An upstream PR adding
`incident_type`/`incident_types` to `heimweh/go-pagerduty` should be filed as a tracked follow-up
before anyone next re-vendors this dependency.

## Plan reference

This PR was implemented against `__agent-runtime/PLAN.md`. See that file for the full contract,
and `__agent-runtime/REVIEW.md` for the final review notes (verdict PASS at commit 452593f8).


cjgajard:
AI-authored PR — final review verdict: `PASS` after 2 implementation cycles.

**Findings:** 0 blockers, 2 suggestions.

## Suggestions for follow-up

- [SUGGESTION] The vendor patch to `vendor/github.com/heimweh/go-pagerduty` has no accompanying
  `go.mod`/`go.sum` change. This is a deliberate, explicitly-documented decision (see PLAN.md and
  the "Vendor patch fragility" section of the PR description): any future `go mod vendor` or
  `make update-go-pagerduty` run will silently revert the patch. Non-blocking; an upstream PR to
  `heimweh/go-pagerduty` is the tracked follow-up.
- [SUGGESTION] `d.GetOk("incident_types")` in `buildIncidentWorkflowTriggerStruct` treats an empty
  list as "no value," so shrinking `incident_types` to `[]` on an existing `incident_type` trigger
  is not sent to the API and the old list would silently persist server-side. Outside this
  change's stated acceptance criteria (which only cover adding/swapping entries); flagged for
  visibility, not as a blocker.

**Plan reference:** see `__agent-runtime/PLAN.md` for the contract this PR was implemented against.

cjgajard:
Follow-up fix pushed: `incident_types` was previously silently ignored when shrunk to an empty list (`d.GetOk` treats an empty list as unset). Now uses `d.Get`, and the vendored `IncidentTypes` field is a `*[]string` so `omitempty` only drops it for non-`incident_type` triggers while an explicit empty list still sends `[]`. Added a test step covering the shrink-to-empty case. Mirrored upstream at heimweh/go-pagerduty#169.

cjgajard:
Found and fixed a real bug via live `make testacc`: the API rejects `is_subscribed_to_all_services` outright (400) for `incident_type` triggers — undocumented in the OpenAPI spec, which claimed no restriction. Fixed the same way as `incident_types`: vendored `SubscribedToAllServices` is now `*bool` so `omitempty` can omit it regardless of its boolean value, it's never sent for `incident_type` triggers, and `validateIncidentWorkflowTrigger` now requires it be `false` in config for that type (clear plan-time error instead of a 400 at apply). Also fixed the acceptance tests to stop creating throwaway `pagerduty_incident_type` resources (they can't be deleted via the API, and that resource isn't even registered in this package's SDKv2-only test provider) — they now reference pre-existing IDs via `PAGERDUTY_ACC_INCIDENT_TYPE_ID`/`PAGERDUTY_ACC_INCIDENT_TYPE_ID_UPDATED`. Mirrored upstream at heimweh/go-pagerduty#169.

cjgajard:
Reverted the `SubscribedToAllServices *bool` change: kept it as a plain `bool` to match `heimweh/go-pagerduty` main. `validateIncidentWorkflowTrigger` already requires `subscribed_to_all_services = false` in config for `incident_type` triggers, and the vendored client's existing `omitempty` tag on the plain bool already drops a `false` value from the request — no pointer indirection needed.

cjgajard:
Two more live-API findings, fixed:
1. `permissions` is also rejected outright for `incident_type` triggers. `permissions` is Optional+Computed, so a computed default written to state by an earlier apply gets picked up by `d.GetOk` on later updates even when config never mentions it. Gated the build on `triggerType` instead of trusting `GetOk` alone.
2. `incident_types` must contain at least 1 element — shrinking it to `[]` isn't a supported operation at all (400 "must contain at least 1 items"), contradicting the earlier shrink-to-empty fix/test. Reverted the vendored `IncidentTypes` field back to plain `[]string` (the pointer trick is unneeded once empty is unreachable) and added a plan-time guard rejecting an empty list for `incident_type`, replacing the shrink-to-empty test with an `ExpectError` test. Mirrored the vendor revert upstream at heimweh/go-pagerduty#169.

cjgajard:
```
+go test ./pagerduty -run TestAccPagerDutyIncidentWorkflowTrigger_ -v
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_import
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_import (19.49s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_import_incidentType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_import_incidentType (10.72s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BadType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BadType (0.44s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithManualType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithManualType (0.35s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithIncidentType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ConditionWithIncidentType (0.35s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_SubscribedToAllWithIncidentType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_SubscribedToAllWithIncidentType (0.35s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_SubscribedToAllWithInvalidServices
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_SubscribedToAllWithInvalidServices (0.34s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BasicManual
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BasicManual (18.70s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BasicIncidentType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BasicIncidentType (17.19s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_IncidentTypeWithEmptyIncidentTypes
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_IncidentTypeWithEmptyIncidentTypes (0.43s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_BasicConditionalAllServices
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_BasicConditionalAllServices (23.24s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ManualWithTeamPermissions
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ManualWithTeamPermissions (36.55s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_ChangeTypeCausesReplace
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_ChangeTypeCausesReplace (23.41s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_CannotChangeType
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_CannotChangeType (22.35s)
=== RUN   TestAccPagerDutyIncidentWorkflowTrigger_UpdateToEmptyCondition
--- PASS: TestAccPagerDutyIncidentWorkflowTrigger_UpdateToEmptyCondition (28.64s)
PASS
ok  	github.com/PagerDuty/terraform-provider-pagerduty/pagerduty	203.387s
```

cjgajard:
Depends on https://github.com/heimweh/go-pagerduty/pull/169

